PR #3, opened 2018-01-20
WIP: Add a Servo tree.

emilio:
Still a few things broken re. generated files, and of course crates.io
dependencies and such, but this does the basic work.

### DIFF
--- a/config.json
+++ b/config.json
@@ -41,6 +41,16 @@
       "objdir_path": "$WORKING/whatwg-html/objdir",
       "codesearch_path": "$WORKING/whatwg-html/livegrep.idx",
       "codesearch_port": 8084
+    },
+
+    "servo": {
+      "index_path": "$WORKING/servo",
+      "files_path": "$WORKING/servo/servo",
+      "git_path": "$WORKING/servo/servo",
+      "git_blame_path": "$WORKING/servo/blame",
+      "objdir_path": "$WORKING/servo/servo/target",
+      "codesearch_path": "$WORKING/servo/livegrep.idx",
+      "codesearch_port": 8085
     }
   }
 }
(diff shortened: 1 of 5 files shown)


emilio:
cc @jdm, would you be able to help out taking it from here?

emilio:
(I can also on my free time of course, just curious)

jdm:
Thanks for getting it this far!

staktrace:
Possibly a blocker: https://github.com/rust-lang/rls/issues/1378

asutherland:
Is this still something we're interested in?  I do assume the technical plan here would depend on the rough plan to somehow do stuff inside of rustc similar to how the clang indexer plugin works.

jdm:
I would be really pleased if it existed, but I'd only encourage people to work on making it happen if they're personally invested.

asutherland:
I think in general we're trying to move in a direction where taskcluster (or some other CI) tries to do most of the compilation indexing legwork and the searchfox AWS indexing jobs download those artifacts and then do the indexing, etc.  Does servo have a CI setup (maybe on taskcluster) where the "cargo check" steps could run and generate the save-analysis artifacts?  (And then in the future when analysis data can be generated directly by rustc/lib-ified rustc, that would happen in that CI job.  Right now the searchfox AWS indexing jobs do perform the conversion step from save-analysis... could also happen in the upstream job, but introduces a dependency on the build process, as it would have to check out the mozsearch repo somehow or use a (new) tooltool job/etc.)

In terms of the RAM issue :kats pointed out, the good news there at least is the actual indexer machines we use (m5d.2xlarge) have 32 GiB of RAM, so that hopefully shouldn't be a blocker itself.

jdm:
We're on taskcluster at the moment, but actively moving away from it. It could theoretically be done in a github action, though.

asutherland:
I'm going to close this since this is bit-rotted enough (and will become even more bit-rotted soon) that someone would probably want to start over.

PR #4, opened 2018-02-05
Allow job types of build-linux64/debug as valid c-c target jobs.

staktrace:
Apparently at some point recently the job type of linux64 debug builds on c-c changed. This updates the script to allow either the old or the new job type.

(This fixes the problem for me locally but I haven't tested a deployment with it yet)


staktrace:
Going to merge this now so that today's normal indexing run (which will start in ~2.5 hours) picks this up. It would definitely fail without this, so even if it fails with this at least it won't be any worse. And hopefully this patch works and the run will pass.

PR #7, opened 2018-03-31
Update comm-central to use git repo

staktrace:
This fixes bug 1446284. Tested locally and it works as expected. I kicked off a push with this to dev to verify it works fine in production.

I made the tarballs for the git repo and blame repo and pushed them manually to bootstrap the process, but the scripts will continue to update them like they do for gecko-dev.


config.json
```diff
@@ -27,6 +27,8 @@
     "comm-central": {
       "index_path": "$WORKING/comm-central",
       "files_path": "$WORKING/comm-central/comm-central",
+      "git_path": "$WORKING/comm-central/comm-central",
+      "git_blame_path": "$WORKING/comm-central/comm-central-blame",
       "hg_root": "https://hg.mozilla.org/comm-central",
```
emilio: Hmm, does this need to go away?

staktrace: hg_root is still used for "raw" links, and to get icons in the directory listing at https://github.com/mozsearch/mozsearch/blob/07726ffa17fbc9fc4ec116603bdcbea7b31d7aa6/output-dir.js#L99 for example. So it still provides some value.

PR #12, opened 2018-09-08
Use taskcluster artifacts

staktrace:
Implementation is not that complicated, but it's a conceptual change hence the double review request


mozilla-central/build
```diff
@@ -21,7 +21,12 @@ mkdir -p $OBJDIR
 $MOZSEARCH_PATH/scripts/indexer-setup.py >> $INDEX_ROOT/mozconfig
 
 cd $FILES_ROOT
-MOZCONFIG=$INDEX_ROOT/mozconfig ./mach build
+if [ -f $INDEX_ROOT/target.mozsearch-index.zip ]
```
emilio: nit: I'd use quotes consistently here.

mozilla-central/setup
```diff
@@ -46,7 +49,102 @@ pushd $GIT_ROOT
...
+# Edit this to e.g. try.revision.ee64db93dcc149da9313460317257b8c42eec5b2 or whatever
+# to test other revisions. If you do this, also override INDEXED_GIT_REV below to some sane
+# git-equivalent of the revision because most likely it won't find anything in the mapfile.
+# This defaults to the most recent searchfox indexing run for mozilla-central.
+REVISION=mozilla-central.latest
+
+pushd $INDEX_ROOT
+wget -nv https://index.taskcluster.net/v1/task/gecko.v2.$REVISION.firefox.linux64-searchfox-debug/artifacts/public/build/target.json
+INDEXED_HG_REV=$(python $MOZSEARCH_PATH/scripts/read-json.py target.json moz_source_stamp)
+INDEXED_GIT_REV=$(awk -v REV=$INDEXED_HG_REV '$2 ~ REV { print $1 }' git_hg.map)
```
emilio: Oh wow, I'm not looking forward to debug this, my awk-foo sucks :)

mozilla-central/setup
```diff
@@ -10,6 +10,9 @@ pushd $INDEX_ROOT
 if [ -d "gecko-dev" ]
 then
     echo "Found pre-existing gecko-dev folder; skipping re-download."
+    pushd $INDEX_ROOT/gecko-dev
```
emilio: nit: I think pushd gecko-dev should be enough, and is consistent with the if condition.

mozilla-central/setup
```diff
@@ -10,6 +10,9 @@ pushd $INDEX_ROOT
...
+    git checkout master
```
emilio: Why is this needed?

emilio: (And why don't we need to `fetch` / `pull`?

mozilla-central/setup
```diff
@@ -46,7 +49,102 @@ pushd $GIT_ROOT
...
+    # in the generated-files tarball that we get from taskcluster. Most of these
+    # missing files are dummy unified build files that we don't care about, and
+    # we can just delete those.
+    # If there are other such cases, then we'll get zero-byte files generated by
+    # output-file.rs since it won't be able to find the source file corresponding to
+    # the analysis file. In those cases we should ensure the generated file is
+    # included in the target.generated-files.tar.gz tarball that comes out of the
+    # taskcluster indexing job. Bug 1440879 can be used as a guide.
+    pushd $INDEX_ROOT/analysis/__GENERATED__
+    find . -type f -name "Unified*" |
```
emilio: It's a bit unfortunate to hardcode this, but I guess the build system doesn't have a list of unified objects? Looks like it shouldn't be too hard to generate such a list, maybe worth filing a bug?

mozilla-central/build
staktrace: Done

mozilla-central/setup
staktrace: basic awk is well worth learning, I find it super useful :)

mozilla-central/setup
staktrace: I don't know if there's much benefit to doing that. I could file a bug but chances are it will never get fixed. A better change might be to strip out the analysis files for the Unified* files from the analysis tarball.

mozilla-central/setup
amccreight: Could you add some of this explanation in a comment here?

mozilla-central/setup
amccreight: What happens if any of these 3 steps fail?

mozilla-central/setup
```diff
@@ -46,7 +49,102 @@ pushd $GIT_ROOT
...
+# This defaults to the most recent searchfox indexing run for mozilla-central.
+REVISION=mozilla-central.latest
+
+pushd $INDEX_ROOT
+wget -nv https://index.taskcluster.net/v1/task/gecko.v2.$REVISION.firefox.linux64-searchfox-debug/artifacts/public/build/target.json
+INDEXED_HG_REV=$(python $MOZSEARCH_PATH/scripts/read-json.py target.json moz_source_stamp)
+INDEXED_GIT_REV=$(awk -v REV=$INDEXED_HG_REV '$2 ~ REV { print $1 }' git_hg.map)
+
+# If INDEXED_GIT_REV gets set, that means the gecko-dev repo includes the code for which
+# The taskcluster indexing job ran. So we can download the result of that indexing
```
amccreight: micronit: "the" not "The"

mozilla-central/setup
```diff
@@ -46,7 +49,102 @@ pushd $GIT_ROOT
...
+# job and use it. Otherwise, it means gecko-dev is lagging behind the canonical hg
```
amccreight: It sounds like you are saying there is a race between the hg-to-git mirror gecko-dev and the SearchFox taskcluster build. If the taskcluster build wins the race, it sounds like we fall back to the current behavior of doing a "local" clang build? Should there be some output in the log indicating which thing happened, or is it clear enough from all of the other log spew from the various steps? How often do we expect to fall back to the non-taskcluster build? Do we have any way of knowing if it starts happening very frequently? It seems like that would be bad (though not bad enough that we'd want SearchFox to break entirely instead). If we're only rarely using the "local" build path, how are we going to ensure that it keeps working?

mozilla-central/setup
```diff
@@ -46,7 +49,102 @@ pushd $GIT_ROOT
...
+# as this file, which detects this scenario by the absence of the
+# target.mozsearch-index.zip file.
+
+if [ -n "$INDEXED_GIT_REV" ]; then
+    pushd $GIT_ROOT
+    # Since we're using a prebuilt index at a particular rev, sync our local repo to the same rev
+    git checkout $INDEXED_GIT_REV
+    popd
+
+    # Download the C++ analysis, rust save-analysis files, and generated sources tarballs.
```
amccreight: micronit: Rust

amccreight:
What do you mean by "it is a conceptual change"? Could you describe what the old and new concepts are? Perhaps in the commit message.

mozilla-central/setup
staktrace: Done

mozilla-central/setup
staktrace: If the wget or python command fail the script will error out (nonzero return value) and the indexing run will fail. I think this is desirable because these commands shouldn't fail. If the hg rev is not found in the mapfile the awk script will exit with zero but won't produce output so we'll skip downloading the tarballs and fall back to doing a full build. There is a comment to this effect just below.

mozilla-central/setup
staktrace: Fixed

mozilla-central/setup
staktrace: > It sounds like you are saying there is a race between the hg-to-git mirror gecko-dev and the SearchFox taskcluster build. If the taskcluster build wins the race, it sounds like we fall back to the current behavior of doing a "local" clang build?

Yes.

> Should there be some output in the log indicating which thing happened, or is it clear enough from all of the other log spew from the various steps?

It will be clear from the log spew, but I do want to add a "warning" mechanism where it will send an email to the mailing list (but still complete the indexing successfully) in this case and other similar cases. I'm dling that separately though.

>  How often do we expect to fall back to the non-taskcluster build?

Not very often. These days the git mirror is pretty good at staying updated.

> Do we have any way of knowing if it starts happening very frequently?

With just this patch you'd have to manually check the log. With the warning mechanism I described above we'd get notified by email every time it happens so it would be easy to keep an eye out for it.

> If we're only rarely using the "local" build path, how are we going to ensure that it keeps working?

This is a potential problem, yes. But I'm hoping that the set of build dependencies doesn't expand that often that it's a major concern. One thing I want to do that will help here is to run `mach bootstrap` before `mach build` in the `build` script which should do a better job of ensuring the "local" build path doesn't break.

staktrace:
While expanding the commit message I realized that as part of the conceptual change the analysis file format becomes part of the "API" in that it's included in the taskcluster artifacts. I put that in the commit message too.

staktrace:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1490159 and https://bugzilla.mozilla.org/show_bug.cgi?id=1490160

PR #17, opened 2018-09-15
Bug 1490160 - Run mach bootstrap before building to ensure build requ…

staktrace:
…irements are satisfied.


staktrace:
Yeah, otherwise it has the interactive prompt asking for which product you want to bootstrap for.

PR #18, opened 2018-09-15
Re-compress the gecko-dev and gecko-blame repos periodically.

staktrace:
Running `git gc` on gecko-dev brought the size down from ~11G
to ~6G and gecko-blame shrunk from ~14G to ~1.7G. This reduces
indexer runtime and data transfer noticeably. This patch
monitors the size of the tarballs and re-compresses the repos
if they get too big.


staktrace:
It can actually be done with bash itself: $((10 * 1000 * 1000 * 1000)) which is perhaps better than a long string of zeros. I've updated the patch to use this along with a named variable which makes it clearer.

PR #19, opened 2018-09-16
Silence uninteresting file listing that goes into the log when unzipp…

staktrace:
…ing stuff


staktrace:
Having to page past all the uninteresting "inflating <file>" messages is annoying when looking at logfiles. This is pretty trivial so I'll just merge it.

PR #22, opened 2018-10-15
Update config.json with GitHub repos.

emilio:
We need https://github.com/mozsearch/mozsearch/pull/159 for this.

This fixes all links except the nss one because I don't know which git repo
we're using. I've verified that all git links are working now except that one
which still points to gecko-dev.


emilio:
@franziskuskiefer do you happen to know which Git repo is searchfox using? I tried nss-dev/nss but that didn't seem to be the same.

franziskuskiefer:
hm, `nss-dev/nss` is the only git mirror of `projects/nss` I know of.

staktrace:
@emilio It is using `nss-dev/nss`, why do you think that's not right?

PR #24, opened 2018-10-31
Windows

staktrace:
Reading the diff on the first patch is likely not that useful - it's better to just read the new version and see if it makes sense. The old code downloaded the Linux artifacts, deployed them to the right places, and then merged the mac stuff into it. The new code makes this a bit more generic by looping over the platforms.

This (second patch specifically) depends on https://github.com/mozsearch/mozsearch/pull/162

### DIFF
--- a/mozilla-central/setup
+++ b/mozilla-central/setup
@@ -90,7 +90,7 @@ if [ -n "$INDEXED_GIT_REV" ]; then
     rm -f analysis-files.list analysis-dirs.list
 
     # Download the artifacts from taskcluster for each platform that we're indexing
-    for PLATFORM in linux64 macosx64; do
+    for PLATFORM in linux64 macosx64 win64; do
         date
 
         # First check that the latest artifacts for the platform match the revision we want, otherwise emit a warning and skip it
--- a/mozilla-central/setup
+++ b/mozilla-central/setup
@@ -65,7 +65,7 @@ function move_file {
 REVISION=mozilla-central.latest
 
 pushd $INDEX_ROOT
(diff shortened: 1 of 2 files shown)


mozilla-central/setup
```diff
@@ -77,135 +77,140 @@ INDEXED_GIT_REV=$(awk -v REV=$INDEXED_HG_REV '$2 ~ REV { print $1 }' git_hg.map)
...
-    # the analysis file. The generated-files tarball only has one copy, so we manually
-    # make the other copy to make things match up. The best would be if the gecko build
-    # system didn't actually produce two copies of these files.
-    cp $INDEX_ROOT/objdir/buildid.h $INDEX_ROOT/objdir/dist/include/buildid.h
-    cp $INDEX_ROOT/objdir/mozilla-config.h $INDEX_ROOT/objdir/dist/include/mozilla-config.h
-
-    date
+    # Download the artifacts from taskcluster for each platform that we're indexing
+    for PLATFORM in linux64 macosx64; do
+        date
```
amccreight: Maybe print the platform, too?

amccreight:
Be sure to announce this on dev-platform when it is up and running. I'm sure some people who deal with Windows only code will be glad to know that it works there, now, and they can file bugs on issues they see. Thanks for working on this. It'll be great to have all of this code indexed.

mozilla-central/setup
staktrace: Updated patch to print the platform. I did it after the revision check, before downloading the artifacts.

staktrace:
Yup, I'll send an email to dev-platform after (fingers crossed) the indexing run tomorrow picks up all the windows code.

PR #25, opened 2018-11-09
Improve handling of generated sources

staktrace:
Instead of preferring generated sources from some platforms, compare
all the platform-specific variants of the generated files. If they match,
then keep a single copy of the file as before. If they don't match, put
each platform variant into a platform-specific subdir of the objdir/ and
under `analysis/__GENERATED__/`.

### DIFF
--- a/mozilla-central/setup
+++ b/mozilla-central/setup
@@ -58,6 +58,36 @@ function move_file {
     mv "$1" "$2"
 }
 
+# Check that all the files (provided as arguments) are the same, after normalizing
+# Windows line endings and paths to UNIX. At least one argument must be provided.
+# If all the files match, the name of the first one is echo'd, otherwise the empty
+# string is echo'd.
+function check_all_same {
+    if [ $# -eq 0 ]; then
+        # At least one arg must be provided
+        return 1;
+    fi
+    FIRSTFILE=$1; shift;
+    # Normalize to UNIX in-place. The z: absolute path comes from the taskcluster
+    # working directory for Windows.
+    dos2unix --quiet "$FIRSTFILE"
+    sed --in-place -e "s#z:/task_[0-9]*/#/builds/worker/workspace/#g" "$FIRSTFILE"
+    while [ $# -gt 0 ]; do
+        NEXTFILE=$1; shift;
+        dos2unix --quiet "$NEXTFILE"
+        sed --in-place -e "s#z:/task_[0-9]*/#/builds/worker/workspace/#g" "$NEXTFILE"
+        cmp --quiet "$FIRSTFILE" "$NEXTFILE"
+        if [ $? -ne 0 ]; then
+            # Files aren't the same, echo empty string
+            echo ""
+            return 0;
+        fi
+    done
+    # All files are the same, echo any one of them
+    echo "$FIRSTFILE"
+    return 0;
+}
+
 # Edit this to e.g. try.revision.ee64db93dcc149da9313460317257b8c42eec5b2 or whatever
 # to test other revisions. If you do this, also override INDEXED_GIT_REV below to some sane
 # git-equivalent of the revision because most likely it won't find anything in the mapfile.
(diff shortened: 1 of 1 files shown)


mozilla-central/setup
```diff
@@ -58,6 +58,36 @@ function move_file {
     mv "$1" "$2"
 }
 
+# Checks that all the files (provided as arguments) are the same, after normalizing
```
amccreight: "Checks" instead of "check"?

mozilla-central/setup
```diff
@@ -153,48 +183,69 @@ if [ -n "$INDEXED_GIT_REV" ]; then
...
+        find . -not \( -name __GENERATED__ -prune \) -type d >> ../analysis-dirs.list
+        find . -not \( -name __GENERATED__ -prune \) -type f >> ../analysis-files.list
         popd
 
     done    # end PLATFORM loop
 
     date
 
+    # Special case: xptdata.cpp is a giant file and is different for each platform, but
+    # the differences are not particularly relevant so let's just keep the linux one.
```
amccreight: Nit: Linux.

mozilla-central/setup
```diff
@@ -153,48 +183,69 @@ if [ -n "$INDEXED_GIT_REV" ]; then
...
+    rm -f              generated-macosx64/xpcom/reflect/xptinfo/xptdata.cpp
+    rm -f                 generated-win64/xpcom/reflect/xptinfo/xptdata.cpp
+    rm -f analysis-macosx64/__GENERATED__/xpcom/reflect/xptinfo/xptdata.cpp
+    rm -f    analysis-win64/__GENERATED__/xpcom/reflect/xptinfo/xptdata.cpp
+
+    # For each generated file, if all platforms generated the same thing (or didn't
+    # generate the file at all), copy it to the merged objdir.
```
amccreight: What sort of files are only generated on some platforms? It feels like it could be confusing but maybe it is obvious enough from the file.

mozilla-central/setup
```diff
@@ -153,48 +183,69 @@ if [ -n "$INDEXED_GIT_REV" ]; then
...
+    set +x  # Turn off echoing of commands and output only relevant things to avoid bloating logfiles
+    sort generated-files.list | uniq |
+    while read GENERATED_FILE; do
+        ALL_SAME_AS=$(check_all_same generated-*/$GENERATED_FILE)
+        if [ "$ALL_SAME_AS" != "" ]; then
+            echo "Generated file $GENERATED_FILE was identical across platforms"
```
amccreight: Maybe this should end with "across platforms where it was created" to spell out the possibility that it existed on some platforms and not others?

mozilla-central/setup
```diff
@@ -153,48 +183,69 @@ if [ -n "$INDEXED_GIT_REV" ]; then
...
+            MERGED_ANALYSIS="analysis/__GENERATED__/$GENERATED_FILE"
+            mkdir -p "$(dirname $MERGED_ANALYSIS)"
+            RUST_LOG=info $MOZSEARCH_PATH/tools/target/release/merge-analyses analysis-*/__GENERATED__/$GENERATED_FILE > $MERGED_ANALYSIS
+            continue;
+        fi
+
+        # The generated file was not the same across all platforms, so
+        # put the different versions in __$PLATFORM__ subfolders.
+        for PLATFORM in linux64 macosx64 win64; do
+            if [ -f "generated-$PLATFORM/$GENERATED_FILE" ]; then
```
amccreight: This loop might look a little better with an "early continue" style. I don't know if that's not the style in Bash or whatever.

mozilla-central/setup
```diff
@@ -153,48 +183,69 @@ if [ -n "$INDEXED_GIT_REV" ]; then
...
+                    move_file "analysis-$PLATFORM/__GENERATED__/$GENERATED_FILE" "analysis/__GENERATED__/__${PLATFORM}__/$GENERATED_FILE"
+                fi
+            fi
+        done
+    done
+    set -x
+
+    # Throw away any leftover per-platform generated files, and the analysis data
+    # for generated files. The above loop should have extracted all the useful
+    # information from thees folders into the objdir/ and analysis/__GENERATED__/
```
amccreight: nit: "thees"

mozilla-central/setup
staktrace: There's some stuff like `gfx/layers/CompositorD3D11Shaders.h` and `accessible/interfaces/ia2/Accessible2.h` which is windows-only. And stuff like `build/unix/elfhack/inject/x86_64.c` is Linux-only. It is mostly obvious from the file in question.

staktrace:
Addressed all review comments, thanks!

PR #26, opened 2018-11-15
Bug 1505869 - Use the distinclude map to improve analysis for files c…

staktrace:
…opied into dist/include.

On Windows, files are copied, rather than symlinked, into dist/include.
This can result in the analysis data ending up in the
`__GENERATED__/dist/include` folder when it should really be at the path
corresponding to the original source location of the header. This patch
uses the distinclude mapfile to correct that.

### DIFF
--- a/mozilla-central/setup
+++ b/mozilla-central/setup
@@ -152,6 +152,19 @@ if [ -n "$INDEXED_GIT_REV" ]; then
         mkdir -p generated-$PLATFORM
         tar -x -z -C generated-$PLATFORM -f $PLATFORM.generated-files.tar.gz
 
+        # Download the manifest for dist/include entries and normalize away the taskcluster paths
+        rm -f $PLATFORM.distinclude.map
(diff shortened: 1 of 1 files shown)


mozilla-central/setup
```diff
@@ -179,7 +192,55 @@ if [ -n "$INDEXED_GIT_REV" ]; then
             fi
         done
         set -x
+        popd
+
+        date
+
+        # On Windows, analysis for headers ends up in __GENERATED__/dist/include/...
```
amccreight: These comments only talk about Windows, but it sounds like this is done for all platforms? Are there files you know that this helps, or is it only for completeness?

mozilla-central/setup
```diff
@@ -179,7 +192,55 @@ if [ -n "$INDEXED_GIT_REV" ]; then
...
+                fi
+            fi
+        done
+        set -x
+        popd
+
+        date
+
         # Also drop any directories that got emptied as a result
+        pushd analysis-$PLATFORM/__GENERATED__
```
amccreight: Why is this change needed?

mozilla-central/setup
staktrace: It only helps on windows, but I did it for all platforms for completeness.

PR #29, opened 2019-02-27
Bug 1500941 - Add android-armv7 indexing

staktrace:
Needs bug 1530016 to merge to m-c before merging this.

### DIFF
--- a/mozilla-central/setup
+++ b/mozilla-central/setup
@@ -122,7 +122,7 @@ if [ -n "$INDEXED_GIT_REV" ]; then
     rm -f generated-files.list analysis-files.list analysis-dirs.list
 
     # Download the artifacts from taskcluster for each platform that we're indexing
-    for PLATFORM in linux64 macosx64 win64; do
+    for PLATFORM in linux64 macosx64 win64 android-armv7; do
         date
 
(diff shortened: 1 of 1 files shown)


mozilla-central/setup
```diff
@@ -266,10 +266,12 @@ if [ -n "$INDEXED_GIT_REV" ]; then
 
     # Special case: xptdata.cpp is a giant file and is different for each platform, but
     # the differences are not particularly relevant so let's just keep the Linux one.
-    rm -f              generated-macosx64/xpcom/reflect/xptinfo/xptdata.cpp
-    rm -f                 generated-win64/xpcom/reflect/xptinfo/xptdata.cpp
-    rm -f analysis-macosx64/__GENERATED__/xpcom/reflect/xptinfo/xptdata.cpp
-    rm -f    analysis-win64/__GENERATED__/xpcom/reflect/xptinfo/xptdata.cpp
+    rm -f                   generated-macosx64/xpcom/reflect/xptinfo/xptdata.cpp
```
emilio: Maybe we should write this as a loop?

staktrace: Good point, fixed. Not sure why I didn't think of that...

PR #31, opened 2019-03-10
Add a mozilla-releases.json config file

staktrace:
These patches add mozilla-{beta,release,esr60} repos, but do so in a new config file. This will allow us to index these trees on a separate indexer/webserver instance than the one that does the existing repos listed in config.json. See https://bugzilla.mozilla.org/show_bug.cgi?id=1282123#c2 for some context.

These patches have no dependencies in that they should be able to land without breaking anything.

### DIFF
--- a/mozilla-release/build
+++ b/mozilla-release/build
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+# Don't do anything here. Once bug 1533801 merges to release
+# we can modify the setup file to pull artifacts from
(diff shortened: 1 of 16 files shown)


mozilla-release/build
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
```
emilio: So we don't want to get any indexing for these for now? It sounds fine, just maybe worth noting in a comment?

staktrace: Added comments in all the new `build` files. I do plan on adding Rust/C++ indexing for mozilla-beta and mozilla-release once bug 1533801 merges to those branches and we can pull the artifacts from taskcluster.

PR #33, opened 2019-03-13
Add blame to mozilla release repos

staktrace:
This gets blame working on the mozilla release repos. The mozilla-central repo is still responsible for updating and uploading the gecko-dev and gecko-blame tarballs to S3 daily, but the release repos will do pull updates for their own branches just in case. Those updates should be no-ops in the common case when everything is working normally.


staktrace:
Thanks for the quick reviews, as always!

PR #39, opened 2019-04-17
Bug 1341203 - Add DXR links for various repos

staktrace:
DXR for comm-central is broken so I didn't include that one


staktrace:
(This requires mozsearch/mozsearch#201 to actually do anything useful)

PR #8, opened 2018-07-29
Export the Rust libstd source to the objdir.

emilio:
This is the required change for https://github.com/mozsearch/mozsearch/pull/123
to work on mozilla-central.

PR #40, opened 2019-04-21
Bug 1545968 - Download and unpack the rust stdlib src and analysis data.

staktrace:
This obsoletes #8. Now we get the src/analysis data from taskcluster rather than installing it on the indexer instance. This ensures that the data we get is in sync with the anaylsis data for the rest of m-c.

This depends on the m-c patches on bug 1545968 which are pending review.

### DIFF
--- a/shared/process-tc-artifacts.sh
+++ b/shared/process-tc-artifacts.sh
@@ -27,6 +27,22 @@ unzip -q $PLATFORM.mozsearch-index.zip -d analysis-$PLATFORM
 # rust-indexer.rs tool will take care of combining all the analysis files correctly.
 unzip -q $PLATFORM.mozsearch-rust.zip -d objdir
 
+# If INDEX_RUSTLIB was set to "yes" during fetch-tc-artifacts.sh, then
+# we'll have per-platform rustlib src/analysis data in zip files, so
+# let's unpack those.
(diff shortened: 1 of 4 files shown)


shared/process-tc-artifacts.sh
```diff
@@ -27,6 +27,17 @@ unzip -q $PLATFORM.mozsearch-index.zip -d analysis-$PLATFORM
...
 
+# These zips have a rustlib top-level folder and then contain the
+# rust stdlib src (all the zips have the same src tree) and analysis
+# data for the platform. We unpack all the zips into the same
+# destination folder, implicitly merging them. And we copy the stdlib
+# src tree into the objdir and (ab)use the generated files machinery.
+unzip -qn $PLATFORM.mozsearch-rust-stdlib.zip -d .
+if [ ! -d "objdir/__RUST__" ]; then
+    mkdir -p "objdir/__RUST__"
+    cp -Rn rustlib/src/rust/src objdir/__RUST__/
```
emilio: Do we want to do this conditionally? I assume this is not a problem when m-c updates their rust version?

staktrace: I don't expect it to be a problem when m-c updates rust, because it's still the `repack_rust.py` script that builds the rust toolchain artifacts, and that's where we patched in the code to include the stdlib src and analysis data. So it should Just Work (TM).

PR #45, opened 2019-05-14
Bug 1551082 - Add some docs for mozsearch-mozilla

staktrace:
I wrote some basic stuff but couldn't think of much else that would be useful to say. If you have suggestions on what else to put in here I'm all ears. If not I can always add more stuff later as needed.


asutherland:
I think this provides really excellent context as to what's going on in this repo!

staktrace:
Thanks! :)

PR #49, opened 2019-07-16
Bug 1565630 - Add Rust/C++ indexing to ESR68.

staktrace:

PR #50, opened 2019-07-16
Index rustlib

staktrace:
PR includes patch from #49 because it applies on top of that.


asutherland:
Hm, so apparently I shouldn't have landed that other one first given that it needed to be rebased which seems to confuse github about this PR.

asutherland:
I resolved the conflict and confirmed that the result is the same.  Just cherry-picking the 
341f838  rev and rebasing it would probably the most clean option, but I'm interested to see how this workflow completes, as it is somewhat handy to be able to do this from the web UI and I don't think we're overly concerned about the extra merge commit.

PR #54, opened 2019-09-06
Bug 1567724 - Correct NSS build regression. r=kats

asutherland:
Per my email replying to the searchfox errors from the config.json run, it turns out our NSPR tarball includes configuration data that hardcodes the /index path.  I mentioned a potential symlink mitigation in there, but I've not undertaken it since I think it's better for NSS to just do the hg checkout every time.  This isn't appropriate for repositories with more turnover, but NSPR has very little turnover and the mozilla hg server has optimized bundle magic.

Please feel free to merge this if you're okay with it, as tomorrow's config.json build will be broken without this.

### DIFF
--- a/nss/setup
+++ b/nss/setup
@@ -7,14 +7,20 @@ set -o pipefail # Check all commands in a pipeline
 date
 
 echo Downloading NSPR
+# We used to have a hand-built tar file at nss-nspr.tar.  This included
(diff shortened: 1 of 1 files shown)


asutherland:
Also, this ran fine and everyone should have gotten a warnings email for the branch (nss-builds-weird) because I ran it against "release" in order to update the "release" checkout.

PR #57, opened 2019-10-05
Remove tarball so that there's more disk space.

emilio:
This allows me to get ash without indexed data in the default VM disk size.

We throw it away and re-generate it on upload anyway, so I think this is
harmless.

### DIFF
--- a/shared/fetch-gecko-tarball.sh
+++ b/shared/fetch-gecko-tarball.sh
@@ -33,3 +33,4 @@ if [ ! -f "$WORKING/${TARBALL}.tar" ]; then
 fi
 
 tar xf "$WORKING/${TARBALL}.tar" -C "$DESTDIR"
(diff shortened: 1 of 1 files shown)


asutherland:
I've only been using the VM to run against the test repo as of late.  I've been triggering AWS builds against "dev" (or "asuth") for the real big repo's.  I'm not sure if that should be the expectation of the VM or if we should size it to be able to handle mozilla-central.

emilio:
> IIRC this script gets called multiple times on the releases indexer. Deleting the tarball is going to cause the tarball to get re-downloaded on each invocation instead of just unpacking the one downloaded on the first invocation. So this will result in more bandwidth and slowness.

Does it? We early-exit if the directory we'd create is present already... Oh I guess we do call it with different destinations... I guess that's fair.

PR #58, opened 2019-10-05
Add config for ash.

emilio:


emilio:
I think this should work once the patch in https://bugzilla.mozilla.org/show_bug.cgi?id=1578495 makes it to the ash repo. It works modulo indexing bits locally.

The only thing I had to do manually is creating a branch in the blame repo... @staktrace do you know what the best way to deal with that is? How did you do it for beta / release / etc?

Should I edit the script to check-for-branch-or-create? Or should I just manually upload the repo with the new branch to AWS and keep the script as-is?

shared/fetch-hg-map.sh
```diff
@@ -12,5 +12,11 @@ set -o pipefail # Check all commands in a pipeline
 if [ ! -f "$WORKING/git_hg.map" ]; then
     # We only need "recent" mapfile entries and attempting to download the full mapfile
     # results in a 503 error so we just get the map entries from some fixed recent date.
-    wget -O "${WORKING}/git_hg.map" -nv https://mapper.mozilla-releng.net/gecko-dev/mapfile/since/2019-01-01
+    date=2019-01-01
```
staktrace: Please use uppercase for variable names

shared/fetch-hg-map.sh
```diff
@@ -14,9 +14,15 @@ if [ ! -f "$WORKING/git_hg.map" ]; then
...
+
     # We combine this and the projects branch for convenience, as they share a
     # lot of revisions.
     for tree in gecko-dev gecko-projects; do
-      curl -SsfL https://mapper.mozilla-releng.net/$tree/mapfile/since/$date >> "$WORKING/git_hg.map"
+      curl -SsfL https://mapper.mozilla-releng.net/$tree/mapfile/since/$date >> "$WORKING/git_hg.map.tmp"
     done
+
+    # Remove duplicate entries from the map, which almost halves its size.
+    cat "$WORKING/git_hg.map.tmp" | sort | uniq > "$WORKING/git_hg.map.tmp"
```
staktrace: The destination file here shouldn't have `.tmp`

shared/fetch-hg-map.sh
```diff
@@ -14,9 +14,15 @@ if [ ! -f "$WORKING/git_hg.map" ]; then
     # results in a 503 error so we just get the map entries from some fixed recent date.
     date=2019-01-01
 
+    rm -f "$WORKING/git_hg.map.tmp"
```
staktrace: Not clear why this is needed if you delete it after merging them anyway

staktrace:
Oh also please add bug numbers to commit messages

shared/fetch-hg-map.sh
emilio: Whoopsies...

shared/fetch-hg-map.sh
emilio: Well, just in case there was garbage from a previous run... I don't think it hurts.

emilio:
Addressed the review comments. @staktrace do you know about the question above regarding the branch in the blame repo? Changing to get-or-create may be best, but want your opinion on it before, and I can certainly one-off-push it updated.

staktrace:
When I did the other branches I ran the blame generation manually and uploaded the resulting modified tarball manually. Reason was that I expected it to take a long time to complete and didn't want it to time out the indexer run. That might still be the case here, so it might be safer to do it manually.

emilio:
Blame generation doesn't take a long time for ash (it's just a few commits on top of mozilla-central), but I'll do that anyway.

emilio:
@asutherland sanity-checked/reviewed the bits that @staktrace didn't get to, in particular:

https://github.com/mozsearch/mozsearch-mozilla/pull/58/commits/9e4346e4e49d2328b4ad39196b48aaf54237d5bd / https://github.com/mozsearch/mozsearch-mozilla/pull/58/commits/41cb5b6ec63eaa20ed5fc3a265a825243195001c / https://github.com/mozsearch/mozsearch-mozilla/pull/58/commits/b5081da4b2274d84e349b6a5b15716fe80ab453e

shared/fetch-tc-artifacts.sh
```diff
@@ -26,12 +26,15 @@ REVISION="${REVISION_TREE}.revision.${INDEXED_HG_REV}"
...
-    ${CURL} https://index.taskcluster.net/v1/task/gecko.v2.$REVISION.firefox.$PLATFORM-searchfox-debug/artifacts/public/build/target.json ||
-    (   echo "WARNING: Unable to find analysis for $PLATFORM for hg rev $INDEXED_HG_REV; skipping analysis merge step for this platform." &&
-        continue
-    )
+    TC_PREFIX="https://index.taskcluster.net/v1/task/gecko.v2.${REVISION}.firefox.${PLATFORM}-searchfox-debug/artifacts/public/build"
+    # First check that the searchfox job exists for the platform and revision we want. Otherwise emit a warning and skip it. This
+    # file is small so it's cheap to download as a check that the analysis data for the platform exists.
+    #
+    # Also check for moz_source_stamp, to handle tasks that exists but failed. We rely on this field for resolve-gecko-revs.sh already.
+    if ! (${CURL} "${TC_PREFIX}/target.json" | grep moz_source_stamp); then
```
staktrace: Did you test the failure case here? Because of the `set -e` and `pipefail` options on the script if the grep command fails it will just abort the whole thing, I think.

asutherland: In Slack, Emilio said "Alright, so apparently that error check is just not working at all, as asking for garbage/target.json just returns a valid json file with a ResourceNotFound description".  So I think the idea is that as long as the network connection to taskcluster is good, any errors end up in the JSON rather than as HTTP status codes/etc.

emilio: Yeah, I did (windows builds are busted in ash).

`pipefail` and `set -e` do account for if conditions. You can check this with:

```sh
#!/usr/bin/env bash

set -x # Show commands
set -eu # Errors/undefined vars are fatal
set -o pipefail # Check all commands in a pipeline

for foo in a b c; do
  if ! (echo "bar" | grep foo); then echo "Warning:";
    continue;
  fi
  echo $foo;
done
```

Interestingly, I learned that the `continue` as it was before cannot work at all, because parenthesis in bash effectively create a new environment (so the `continue` is "outside of the loop").

So this works:

```bash
for foo in a b c; do
  false || continue
  echo "should never reach this";
done
```

But this doesn't:

```bash
for foo in a b c; do
  false || (continue)
  echo "should never reach this";
done
```

So there were two reasons why this check wasn't working at all:

 * The json answer was always an http 200, which means that even if the task doesn't exist curl would never return a non-zero result.
 * The error handling code didn't work at all because of the parenthesis.

But anyhow yeah, I did test the error case :)

staktrace: Great, thanks for testing that. I've run into problems before with parentheses spawning subshells but didn't realize it was a problem here, thanks for catching that too :)

PR #61, opened 2019-11-12
Bug 1595492 - Update mapper downloading.

asutherland:
The mapper files are not .tar.bz2 files with paths that look like:
- build/conversion/nss-dev/.hg/git-mapfile
- ./build/conversion/beagle/.hg/git-mapfile (This is the gecko-dev mapping,
  I'm really hoping beagle is a codename and not the wrong source tree!)

Since we don't care about the paths and there's a single file, we can
use the "-O" option to tar to make it extract the contents to stdout.  So
the tar invocation becomes `-xOj` where x is extract and j is bz2
decompression.

I've tested this locally.

### DIFF
--- a/nss/setup
+++ b/nss/setup
@@ -56,8 +56,7 @@ date
 
 echo Downloading git to hg map
 pushd $INDEX_ROOT
-wget -nv https://mapper.mozilla-releng.net/nss/mapfile/full
-mv full git_hg.map
+curl -SsfL https://moz-vcssync.s3-us-west-2.amazonaws.com/mapping/nss-dev/git-mapfile.tar.bz2 | tar -xOj > git_hg.map
 popd
 
 date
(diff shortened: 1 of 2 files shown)


asutherland:
As per https://github.com/mozsearch/mozsearch/pull/253 I'm going to land this directly and ask for post facto review once I've got things happy.  Also, only going to start with the primary config.json run first.

asutherland:
This worked.  Please let me know if you'd like any fix-ups!

asutherland:
You mean the mention of the "mapper releng tool"?

staktrace:
I was referring to [this](https://github.com/mozsearch/mozsearch/blame/master/README.md#L129) - the curl command can be updated to do download/untar/grep the full mapfile at the new url.

asutherland:
Aha! Yes, thanks.  I mis-parsed s/patch/hunk/.  I'll do that.

PR #62, opened 2019-11-13
Bug 1596107 - Temporarily disable "ash" branch after taskcluster migration due to no recent builds

asutherland:


asutherland:
Landing with a=bustage, basically.

PR #64, opened 2019-11-15
Bug 1596846 - Make sure to init new nested submodules as they appear. r=kats

asutherland:
This is for https://bugzilla.mozilla.org/show_bug.cgi?id=1596846 which is about modifying the git repo you created in https://bugzilla.mozilla.org/show_bug.cgi?id=1530798.  I think I inferred the right actions to cargo cult, sans the call to init the nested submodule, but honestly my git submodule "expertise" is a bit rusty.

My intent here is to make sure that if our submodules gain any new submodules, they get properly initialized.  I understand this to be generally safe and correct.  The thing I'm a little concerned about is really the interaction with "--remote" in terms of the nested submodules.  I don't think I'm changing the semantics around that, but it appears like FirefoxReality's submodule is a specific revision to be checked out and my fear would be if we end up just always showing the tip of the `vrb` submodule instead of what the FirefoxReality repo wants checked out.

Of course, the easiest thing may be just to see what happens and ask :Agi on the bug to report back as to whether things seem right or wrong. :)

### DIFF
--- a/mozilla-mobile/setup
+++ b/mozilla-mobile/setup
@@ -22,7 +22,8 @@ date
 
 echo Updating git
 pushd $GIT_ROOT
-git submodule update --recursive --remote
+git submodule update --remote # pull latest moz-mob code
+git submodule update --init --recursive # update nested submodule to desired revision
 git commit --allow-empty -am "Submodule update at $(date)" --author "Searchfox Indexer <searchfox-aws@mozilla.com>"
 popd
 


staktrace:
From my reading of the man page, it seems like this particular change should be fine, but the combination of `--remote` and `--recursive` may not produce desirable results. We might want to split it into two commands:
```
git submodule update --remote # pull latest moz-mob code
git submodule update --init --recursive # update nested submodule to desired revision
```

But anyway I'll r+ this patch since it's orthogonal and the `--init` should be fine.

asutherland:
What you're saying matches up with my concern.  I'm going to update the code to be exactly what you're suggesting.  I'll let Saturday's normal indexing run prove out that everything doesn't explode.

PR #65, opened 2019-11-26
Bug 1593833 - rust third_party indexing support.

asutherland:


shared/process-tc-artifacts.sh
```diff
@@ -13,6 +13,12 @@ fi
 PLATFORM=$1
 INDEX_ROOT=$PWD
 
+# The rust-indexer needs the ability to normalize third-party absolute
+# build paths to be relative to the source directory.  This is currently what
+# taskcluster uses, and you can see other similar normalizations below of
+# $PLATFORM.distinclude.map.
+export BUILT_DIR=/builds/worker/workspace/build/src/
```
staktrace: On Windows doesn't the path start with `z:`?

asutherland: Indeed it does.  I faked myself out on this one while partway through implementing per-platform lookups... the zip extraction had sane-looking paths so I somehow presumed our move to using clang on windows meant we'd started using a more cygwin-looking path structure or that we were cross compiling.  But an example path is `"Z:\\task_1574589562\\build\\src\\third_party\\rust\\euclid\\src\\macros.rs"` which is as you say.

The good news is it looks like std::path::Path can understand windows-style paths even while on linux.  (I have not used rust on Windows as of yet and hadn't ever thought about these potential issues.)

The bad news is that the path's directory name seems to be of the form `task_${Math.floor(Date.now()/1000)}` which is not part of the task id.  The good news is that it's included in the target.mozinfo.json artifact under `topsrcdir`, with that build having the value `"z:/task_1574676011/build/src"` so we can avoid having to guess/hardcode any of this stuff or sprinkle regexps everywhere, although maybe we'll have to download some more stuff.  (localconfig.json also has the path prefix expressed, but doesn't include "src".)

I'll try and quickly get something working based on target.mozinfo.json, although I'm definitely open to suggestions!

asutherland: Hm, and I do see we have the json file using "/" and rust using "\\" so I'll also make sure that the logic normalizes enough but not too much...

staktrace: Another option might be to run something like [this command](https://github.com/mozsearch/mozsearch-mozilla/blob/2bf95a83478eaa489d59845519ae5d8cd186a07e/shared/collapse-generated-files.sh#L37) on the rust analysis files to normalize all the windows paths to Linux

asutherland:
I evolved the approach for this in response to :kats' feedback and https://github.com/mozsearch/mozsearch-mozilla/pull/87 landed for that.

PR #67, opened 2019-12-16
Compute ash blame as part of the mozilla-central setup.

emilio:


emilio:
Testing this locally atm, but it should work :)

asutherland:
(Because right now the the tarball gets updated at https://github.com/mozsearch/mozsearch-mozilla/blob/06cf0aee15aaf20123759c597ed7f08a3e73dd7c/mozilla-central/upload#L38)

staktrace:
Oh good point. Moving the ash indexing to the mozilla-releases indexer instance/config would also solve this, since they start staggered and m-c's upload should run before the releases indexer's setup phase. There are probably other benefits to this as well, such as ash bustage not bring able to break m-c indexing.

asutherland:
Since emilio fixed the platform skipping logic, in general searchfox stability has been great.  (Other than the taskcluster migration blip where there were some mapper hiccups and I needed :peterv to push some stuff, all our breakage I think has been my fault :)

Although putting ash in tier 2 does match priorities, I think in terms of indexer runtime and disk usage it's probably good where it is.  Right now the config.json disk usage is 163G/295G and the mozilla-releases.json is 214G/295G, with ash being 56G of that.

PR #69, opened 2020-02-04
Bug 1613085 - Remove ash indexing.

emilio:


emilio:
Wasn't planning to. My guess is that ash doesn't take much space (it was mostly a bunch of commits on top of central). But if we can do it in the cinnabar transition that's great.

PR #70, opened 2020-02-18
Add Glean to the index

badboy:
Glean is the new Telemetry framework, written in Rust with
platform-specific layers on top (currently Kotlin, Swift and Python are
supported).

/cc @emilio

### DIFF
--- a/glean/upload
+++ b/glean/upload
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+date
+
+echo Uploading glean
+pushd $INDEX_ROOT
+tar cf glean.tar glean
+python $AWS_ROOT/upload.py $INDEX_ROOT/glean.tar searchfox.repositories glean.tar
+rm glean.tar
+popd
+
+date
+
+echo Uploading glean blame
+pushd $INDEX_ROOT
+tar cf glean-blame.tar glean-blame
+python $AWS_ROOT/upload.py $INDEX_ROOT/glean-blame.tar searchfox.repositories glean-blame.tar
+rm glean-blame.tar
+popd
+
+date
--- a/config.json
+++ b/config.json
@@ -72,6 +72,17 @@
       "objdir_path": "$WORKING/mozilla-build/dist",
       "codesearch_path": "$WORKING/mozilla-build/livegrep.idx",
       "codesearch_port": 8087
+    },
+
+    "glean": {
+      "index_path": "$WORKING/glean",
+      "files_path": "$WORKING/glean/glean",
+      "git_path": "$WORKING/glean/glean",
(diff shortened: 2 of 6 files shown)


glean/upload
```diff
@@ -0,0 +1,25 @@
...
+tar cf glean.tar glean
+python $AWS_ROOT/upload.py $INDEX_ROOT/glean.tar searchfox.repositories glean.tar
+rm glean.tar
+popd
+
+date
+
+echo Uploading glean blame
+pushd $INDEX_ROOT
+tar cf glean-blame.tar blame
```
badboy: One thing: should this remove the `target` directory first?

staktrace: Assuming you meant to put this comment up on the part where you `tar` up the `glean.tar` file. And assuming the `target` directory you're referring to is the one generated by `cargo check` in the `build` script. If so, then I think the answer is yes. Although I'm not sure if just running `cargo check` will be sufficient to cause searchfox to actually index the rust code, because the `rust-analyze.sh` script looks for analysis files under the `$OBJDIR` folder. And this patch sets the glean `$OBJDIR` to `$WORKING/glean/dist` which is not where `cargo check` will emit stuff. So really right now the `cargo check` is not doing anything useful.

staktrace:
Hm not sure what happened there, my comment ended up getting posted twice. Sorry about that...

glean/upload
badboy: Ah yes, obviously that comment should have gone on the part above.

Do you have an example how the `cargo check` _should_ work? (I can also work with @emilio in person this week, he promised to have some spare minutes for me)

staktrace: So until now I think the only repos that do rust indexing are mozilla-central based repos, and they run the `cargo build` or `cargo check` steps in CI (on TreeHerder), and then download the analysis files emitted by the rust compiler.

Doing it here as part of the `build` script is possible, but note that it will only run on Linux, and so if your repo contains code that is platform-dependent, then only the Linux branch will get analyzed. If you have a standalone CI process it would be better in the long run to have that emit the rust analysis artifacts and then download them to the searchfox indexer, the way mozilla-central does.

That being said, for now the simplest thing is to just add `--target-dir=../dist` or equivalent to your `cargo check` invocation which will solve both problems (that of having to delete `target/` prior to upload, and that of having the `rust-analyze.sh` script find the analysis data).

I'm currently testing these changes locally and should be updating this PR with a working config shortly. I've also uploaded the necessary tarballs to S3 to bootstrap the repo.

staktrace:
PR updated so that it should work. I'll do a test deployment to double-check.

staktrace:
https://kats.searchfox.org/glean/source has the test deployed (minus the change to /help.html, which I added after I kicked off the test deploy). It seems to be looking good so I'll merge this PR.

staktrace:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1621931

PR #71, opened 2020-02-21
Bug 1607880 - Commit the changes to the superproject before restoring…

staktrace:
… all the submodules to the desired state

### DIFF
--- a/mozilla-mobile/setup
+++ b/mozilla-mobile/setup
@@ -29,8 +29,13 @@ pushd $GIT_ROOT
 # various things were resolved.  For example,
 # https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt#L36
 git submodule update --remote || git submodule update --remote
-git submodule update --init --recursive # update nested submodule to desired revision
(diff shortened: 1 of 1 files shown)


mozilla-mobile/setup
```diff
@@ -29,8 +29,13 @@ pushd $GIT_ROOT
 # various things were resolved.  For example,
 # https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt#L36
 git submodule update --remote || git submodule update --remote
-git submodule update --init --recursive # update nested submodule to desired revision
+# The previous step will update the submodules of this top-level synthetic superproject,
+# and now we want to commit those changes, so that the top-level superproject points
+# to the latest version of each submodule.
+git add *
```
emilio: `git add .` may be slightly better (not requiring shell expansion and so on).

staktrace: I kind of did '*' intentionally to only pick up the top-level submodules, since '.' will pick up everything in subfolders too. It probably doesn't matter much since if there are changes inside the submodules they are technically in separate repos and so shouldn't get picked up but I felt it was safer this way.

PR #73, opened 2020-03-04
Bug 1619363 - Be case-insensitive when normalizing windows z: paths. r=kats

asutherland:
I tested the concept on one of the web-servers but haven't run with it.


asutherland:
Feel free to land if it looks sane, I won't quite make it to my desk before the indexers do their thing.

staktrace:
I'm PTO today so I don't want to merge right now in case it breaks something but I can merge tonight if you don't do it before then.

PR #74, opened 2020-03-05
Bug 1617567 - Move ESR60 to a new mozilla-archived config.

staktrace:
I created a new target group called `mozilla-archived-target`, ran the indexer on this config, and hooked it all up, so it's live. Going to https://searchfox.org/mozilla-esr60/source right now will serve the content from the `mozilla-archived` web-server instance. I'll update docs and add more ESR repos in separate patches.


staktrace:
> Should we be re-numbering mozilla-esr68's codesearch port down 1?

Ah, good point. I notice that config.json also has a number gap (maybe from when ash was removed?) so I'll fix that too.

PR #78, opened 2020-03-07
Bug 1617567 - Have the m-c indexer build blame for more ESRs

staktrace:
Re-landing after manually updating the blame repo.


staktrace:
Same as #76 but now that I've manually added the branches to the blame repo it should work better. Also I fixed the typo from esr32 -> esr31.

PR #87, opened 2020-04-07
Bug 1593833 - Update path normalizations and apply to rust. r=kats

asutherland:


shared/process-tc-artifacts.sh
```diff
@@ -47,8 +47,10 @@ tar -x -z -C generated-$PLATFORM -f $PLATFORM.generated-files.tar.gz
 
 date
 
+NORMALIZE_EXPR="s#/builds/worker/checkouts/gecko/##g;s#z:/task_[0-9]*/build/src/##gI"
+
 # Run the rust analysis here.
-$MOZSEARCH_PATH/scripts/rust-analyze.sh $CONFIG_FILE $TREE_NAME $PLATFORM
+$MOZSEARCH_PATH/scripts/rust-analyze.sh $CONFIG_FILE $TREE_NAME $PLATFORM "$NORMALIZE_EXPR"
```
staktrace, line 53: I'd actually prefer to do the normalization right here if possible, instead of having `rust-analyze.sh` do it. Would that be possible, or are there complications with that?

asutherland: My rationale was that:
- process-tc-artifacts.sh is happening in parallel across all platforms, including the unzipping of the rust artifacts into the objdir
- As part of the changes I made in https://bugzilla.mozilla.org/show_bug.cgi?id=1583031 there's now a bunch of path-logic in https://github.com/mozsearch/mozsearch/blob/master/scripts/rust-analyze.sh and I didn't want to end up duplicating that again in process-tc-artifacts.sh. (And more naive options would perhaps result in the parallel per-platform jobs stomping on each other.)

Perhaps the right thing for us to do is to simplify things by unzipping the rust files into per-platform directories.  The normalization step can then be run naively across the entire directory tree, as can the rust-analyze.sh step.  I think I was perhaps too overly conservative in trying to perturb the existing pipeline when attempting to address bug 1583031.

asutherland:
With `$4` fixed things sorta worked, but I ran into the problem where I'd missed that the rust save paths use actual backslashes and don't normalize to front-slashes like the XPIDL compiler.  So they looked like `"Z:\\task_1586247517\\build\\src\\third_party\\rust\\url\\src\\lib.rs"`.

I attempted to fix this with the most recent set of patches, which included changing some double quotes to single quotes.  Local testing had suggested it should work, but the current state of patches somehow resulted in:
1. Indexing declared success, with the web-server spun up and a success email sent, but...
2. The ~/index dir was empty of everything but the nginx-cache dir.
3. There was no index-log.

Avoiding passing the sed variables through an extra layer of potential unescaping is looking real good right now.  @staktrace Does splitting out the rust save-analysis files into separate directory trees like I propose in the line-item comment above?  (I'm also okay if you want to steal this bug if this is an easier-done-than-described thing.)

shared/process-gecko-analysis.sh
```diff
@@ -51,5 +44,7 @@ sort analysis-files.list | uniq | parallel --halt now,fail=1 "RUST_LOG=info $MOZ
 
 date
 
-# Free up disk space, we don't need these per-platform analysis files any more.
-rm -rf analysis-*
+# We previously would delete the generated-* and analysis-* files and
+# directories at this point because they were no longer necessary for subsequent
+# processing, but this has greatly complicated diagnosing analysis problems,
+# so we now leave them around.
```
staktrace, line 50: There's a chance this will cause the mozilla-releases indexer to run out of disk space. Currently the index volume on that channel is at 72% capacity, and it has three repositories (-beta, -release, -esr68) that will run this processing and leave these files lying around. Might want to check that it doesn't blow up. But note also that we keep the downloaded tarballs, so in case of diagnosing analysis errors, the dev doing the debugging could just re-extract the files from there. Not sure how much of a hassle that would be.

shared/process-tc-artifacts.sh
```diff
@@ -47,8 +64,56 @@ tar -x -z -C generated-$PLATFORM -f $PLATFORM.generated-files.tar.gz
...
+# /builds/worker/checkouts/gecko/obj-arm-unknown-linux-androideabi/dist/xpcrs/rt/nsIChannel.rs
+# Handle this, and do it before we do the source normalization in the next line.
+NORMALIZE_EXPR+=';s#/builds/worker/checkouts/gecko/obj-[-a-zA-Z0-9_]+/#__GENERATED__/#g'
+# Non-windows source paths get normalized off.
+NORMALIZE_EXPR+=';s#/builds/worker/checkouts/gecko/##g'
+# Non-windows build paths get normalized to __GENERATED__.
+NORMALIZE_EXPR+=';s#/builds/worker/workspace/obj-build/#__GENERATED__/#g'
+# I haven't actually seen the same weird source path with "obj-*" under it on
+# Windows yet, but let's proactively assume such a thing will happen and add
+# the following regexp just in case.  But right now nsIChalle.rs on Windows
```
staktrace, line 88: nsIChalle.rs presumably should be nsIChannel.rs?

shared/process-gecko-analysis.sh
asutherland: Keeping the tarballs seems like a good compromise[1] and great idea, thank you.  I hadn't run the disk usage on the leftovers... there's a LOT of leftovers (~5G) per dir... I was thinking they got emptied out more than they do.

1: And arguably better than keeping the directories around since I'm mainly interested in the pre-merge state, not the post-merge state.  And that way they can be extracted from the S3 mount to the local instance mount directly, which is a win over trying to move everything back into place.

shared/process-tc-artifacts.sh
asutherland: Yes, thanks.  I'm really hoping this was a result of me switching back to CUA bindings in VS Code somehow!

shared/process-gecko-analysis.sh
asutherland: Okay, the footer block will now be:
```bash
# Delete the generated-* and analysis-* directories, but retain the tarballs
# for ease of investigation.  (The tarballs are still available from taskcluster
# though, and fetch-tc-artifcats.sh knows how to re-fetch them, so if we don't
# want to waste the space, we could just make it easier to re-fetch them so
# there's a one-liner in mozsearch's `docs/aws.md`.)
rm -rf generated-*/
rm -rf analysis-*/
```

The net change is we leave the tarballs around.  This does represent a deferral of the removal of the `analysis-*/__GENERATED__` sub-trees to this block.  I had eliminated that removal when I had been (tired and) confused and hadn't realized that rust-indexer wasn't generating the nsIChannel.rs analysis output (because I didn't have RUST_LOG=info in process-tc-artifacts.sh), rather than it having been deleted.  This was the objdir/generated (files) confusio that led to the late rename.

Technically that `analysis-*/__GENERATED__` removal could go back, but since the analysis-dirs.list and analysis-files.list explicitly exclude the __GENERATED__ sub-paths and we don't need the disk space at that instance, there's also no harm in the deferral.  So I'm going to leave it as-is, but I don't have strong opinions other than not wanting to risk somehow breaking things.  (The main difference seems like the earlier deletion would resut in a different disk placement for the non-generated analysis merges, but we're not dealing with spinning disks anymore, so disk placement is largely immaterial.)

staktrace: Sounds good to me!

asutherland:
Sometime after mozilla-releases finishes its daily run, I'm going to re-trigger it so we can make sure that there are no surprises for tomorrow.  (I'd re-trigger now and test the indexer reaping, but I realize it's not a holiday for all of Mozilla!)

staktrace:
> Sometime after mozilla-releases finishes its daily run, I'm going to re-trigger it so we can make sure that there are no surprises for tomorrow. (I'd re-trigger now and test the indexer reaping, but I realize it's not a holiday for all of Mozilla!)

I wouldn't object if you did it now. A day of delay on the releases repo isn't the end of the world.

asutherland:
> I wouldn't object if you did it now. A day of delay on the releases repo isn't the end of the world.

Okay, done.  It successfully terminated its predecessor and is doing stuff now.

asutherland:
(And I did it via lambda.  I created a new JustRun config for mozilla-releases.  I'm assuming the python3 switchover resulted in creating new lambda jobs as part of the safe transition of versions.)

staktrace:
> (And I did it via lambda. I created a new JustRun config for mozilla-releases. I'm assuming the python3 switchover resulted in creating new lambda jobs as part of the safe transition of versions.)

I did create new lambda jobs using python3 but you can also trigger locally (assuming you're in python3 virtualenv with latest mozsearch code) and it should work the same.

PR #89, opened 2020-04-29
Bug 1627532 - Switch to rust blame-builder

staktrace:


staktrace:
I did a try run with this on the `kats` channel and it did the right stuff

mozilla-build/setup
```diff
@@ -40,6 +40,6 @@ git pull
 popd
 
 echo Generating blame information
-CINNABAR=1 python $MOZSEARCH_PATH/blame/transform-repo.py $GIT_ROOT $BLAME_ROOT
```
emilio, line 43: You can't miss the `CINNABAR=1` unless I'm missing something, right?

staktrace: The last patch in https://github.com/mozsearch/mozsearch/pull/311/commits made CINNABAR=1 the default. It's still possible to set CINNABAR=0 to explicitly disable cinnabar usage, but that should only be needed if cinnabar is not installed, otherwise we'll just get `None` for hg revs and skip the hg info.

emilio: Ah woops, forgot about it. Looks good then, carry on :)

PR #90, opened 2020-05-03
Pull m-c repos using cinnabar

staktrace:
This makes us pull directly from source-of-truth repos instead of gecko-dev.

### DIFF
--- a/shared/checkout-gecko-repos.sh
+++ b/shared/checkout-gecko-repos.sh
@@ -30,20 +30,15 @@ date
 
 echo Updating git
 pushd $GIT_ROOT
-git fetch origin
-# Fetch projects. Currently unused but makes it easier to index project branches when needed
-git config remote.projects.url || git remote add projects https://github.com/mozilla/gecko-projects.git
-git fetch projects
-# Fetch hg metadata and graft it to the gecko repository using cinnabar. If we want to index
-# project branches we'll want to add the equivalent hg repos and graft metadata from those as
-# well.
+# Fetch mozilla-unified with non-grafted cinnabar, so it will have all the necessary hg metadata.
+# If we need to fetch project branches in the future, we can fetch those also with cinnabar here
 git config remote.cinnabar.url || git remote add cinnabar hg::https://hg.mozilla.org/mozilla-unified
-git config cinnabar.graft true
-git remote update cinnabar
+git config cinnabar.graft false
+git config fetch.prune true
+git fetch cinnabar
 
 # If a try push was specified, pull it in non-graft mode so we actually pull those changes.
 if [ "$REVISION_TREE" == "try" ]; then
-    git config cinnabar.graft false
     git cinnabar fetch hg::https://hg.mozilla.org/try $INDEXED_HG_REV
 fi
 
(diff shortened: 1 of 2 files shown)


shared/checkout-gecko-repos.sh
```diff
@@ -30,20 +30,27 @@ date
...
-git config cinnabar.graft true
-git remote update cinnabar 2> >(grep -v "WARNING Cannot graft" >&2) # Filter stderr to remove warnings we don't care about. Drop once our cinnabar version includes glandium/git-cinnabar#241
+# Fetch the m-c repos that we care about with non-grafted cinnabar, so it will have all the necessary hg metadata.
+# This could be simplified by using mozilla-unified, but currently mozilla-unified is updated with some amount
+# of latency and that can still leave us with stale data. It's better to pull from the individual source-of-truth
+# repos directly. Note we only pull the "default/tip" branch from these remotes, rather than all the random tags
+# and offshoot branches/closed heads that have accumulated in these repos.
+# If we need to fetch project branches in the future, we can fetch those also with cinnabar here.
+# Note that this repo may still have a 'projects' and a 'cinnabar' remote left over that we don't use any more.
+git config remote.central.url || git remote add -t branches/default/tip central hg::https://hg.mozilla.org/mozilla-central
```
emilio, line 40: Do we want central or unified? My thought was that unified was closer to what gecko-dev really contains, and it contains more history.

emilio: We could potentially even just fetch mozilla-unified and use the bookmarks to check out the esrs and such.

asutherland: I read the comment as saying that if we pull from unified we may end up in the state where it doesn't contain the revision that the nightly was built from, a fatal situation analogous to when gecko-dev would break.

emilio: Ah, true, should learn how to read. It'd be nice for unified to be faster I guess, but...

A bit concerned about the blame / history though.

asutherland: It's my understanding that the cinnabar transform was identical in this case between unified and central, because unified is just all of the other repositories smooshed into one repo and so those commits are also identical.  So blame/history are the same either way.  There's just more typing this way.

staktrace: Yeah, what @asutherland said. I would also prefer to use unified, but I definitely ran into cases where it's behind (see https://bugzilla.mozilla.org/show_bug.cgi?id=1625977#c1) and so it could cause failures because we don't pull a rev that taskcluster already has built searchfox data for. And yeah, unified is hash-identical to the source repos so that shouldn't cause any problems with history/blame.

staktrace:
Rebuilding the gecko tarball really just means cloning from mozilla/gecko-dev and then grafting hg metadata from either unified or all the individual repos. After the graft is done the tarball should be equivalent to what we have now, and we can resume pulling from hg.m.o in non-graft mode. Is your concern the history from the CVS days? mozilla/gecko-dev has that, but AFAIK none of the hg.m.o repos (neither m-unified nor m-c) have that, so switching between m-u and m-c doesn't affect that.

staktrace:
I updated the backups of gecko-dev.tar and gecko-blame.tar in the S3 bucket, just in case.

PR #98, opened 2020-05-28
Bug 1641679 - Update gecko-dev branches from the correct remote.

staktrace:
We were updating the branches in gecko-dev from the "origin" remote which we're no longer pulling updates to. I should have just deleted that remote when I was migrating to cinnabar, as that would have caused this line to throw an error which I would have caught. But anyhoo, this should fix it so we update from cinnabar. I'm also manually updating the blame repo and will push it to S3 so tomorrow's indexer don't have to do as much work. I've backed up the current tarballs in case I screw up.


asutherland:
> s/anyhoo/anyhow in the commit message :)

https://www.merriam-webster.com/dictionary/anyhoo

emilio:
Oh, TIL, thanks @asutherland!

PR #100, opened 2020-06-14
Bug 1616362 - Run upload.py as an executable with whatever python it …

staktrace:
…indicates in the shebang line


staktrace:
I missed these files when I did this earlier in #92

PR #101, opened 2020-06-17
Bug 1646069 - Normalize rust build script output paths. r=kats

asutherland:
Per discussion in https://github.com/mozsearch/mozsearch/pull/348 normalize
paths to:

- `__GENERATED__/__linux64__/__RUST_STDLIB__/...` for stdlib
- `__GENERATED__/__linux64__/__RUST_BUILD_SCRIPT__/webrender/shaders.rs` for crate-generated

### DIFF
--- a/shared/process-tc-artifacts.sh
+++ b/shared/process-tc-artifacts.sh
@@ -44,6 +44,63 @@ unzip -q $PLATFORM.mozsearch-rust.zip -d objdir-$PLATFORM
 mkdir -p generated-$PLATFORM
 tar -x -z -C generated-$PLATFORM -f $PLATFORM.generated-files.tar.gz
 
+date
+
+# ### Normalize generated source files by rust build-scripts. ###
+#
+# These will look like:
+#   x86_64-unknown-linux-gnu/debug/build/cranelift-codegen-17ba72a9572695af/out/binemit-x86.rs
+# which can be generalized to:
+#   ${RUST_PLATFORM}/${BUILD_TYPE}/build/${CRATE_NAME}-${CRATE_HASH}/out/${FILE_PATH}
+# and which we want to end up with the following path scheme inside this
+# platform specific directory:
+#   __RUST_BUILD_SCRIPT__/${CRATE_NAME}/${FILE_PATH}
+# which will in the end look like (where __linux64__ varies per platform):
+#   __GENERATED__/__linux64__/__RUST_BUILD_SCRIPT__/${CRATE_NAME}/${FILE_PATH}
+#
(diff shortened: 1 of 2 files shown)


asutherland:
Gonna do a test run now and update the pull request after to add the shaders.rs check.  Inevitably I'll have messed some stuff up, but this covers the broad strokes of the implementation.  I've also pushed an update to https://github.com/mozsearch/mozsearch/pull/348 which is just a cut-down version of these changes.

shared/process-tc-artifacts.sh
```diff
@@ -44,6 +44,63 @@ unzip -q $PLATFORM.mozsearch-rust.zip -d objdir-$PLATFORM
...
+function move_file {
+    mkdir -p "$(dirname $2)"
+    mv "$1" "$2"
+}
+
+# Use sed to get a list of all files that match the patterns we specify above
+# and print out ONLY those files, with each resulting line containing the
+# source path followed by a space followed by the target path.  We will then
+# move the files to their target using the `move_file` helper from above which
+# will create the
```
staktrace, line 83: ... directories as needed? The suspense is killing me!

asutherland:
There end up being 2 meaningful changes necessary:
1. We needed "-E" for the distinclude.map sed transform.  I hadn't added that last time because it wasn't necessary, but that was foolish.
2. The normalization transform had a too-greedy `(.+)` to try and capture the filename, but it was capable of capturing beyond the bounds of the string, resulting in only a single transform being applied.  Whoops!  (This is/was not a problem for the move transform logic which is line-centric.)

I'm re-running again with the latter change imminently.  With just the former change, the files ended up where they were expected, but were missing most of their analysis.

asutherland:
That run worked, giving us https://asuth.searchfox.org/mozilla-central/source/__GENERATED__/__linux64__/__RUST_BUILD_SCRIPT__/webrender/shaders.rs having 1692 lines of analysis just like we get from the existing https://searchfox.org/mozilla-central/source/__GENERATED__/x86_64-unknown-linux-gnu/debug/build/webrender-2a3586aec96fcf3d/out/shaders.rs (although the contents are not identical, presumably owing to the file containing multiple different digests).

shared/process-tc-artifacts.sh
asutherland: You guessed correctly, although perhaps tautologically, as I went with your suggestion here.  I think that's what I was planning to type before my abrupt context switch.

asutherland:
I forgot to add the check.  Going to do that now in a separate PR for OPTIMIZED_SHADERS.

PR #104, opened 2020-07-09
Bug 1616362 - Replace the monolithic per-repo find-repo-files script …

staktrace:
…with only the necessary extension functions.

This removes the old find-repo-files.py script which was shared by all the repos,
and special-cased for comm-central. Now we have smaller files that override
the default behaviour for specific repos that desire to do so.


staktrace:
This goes with https://github.com/mozsearch/mozsearch/pull/352

PR #105, opened 2020-07-16
Bug 1650538 - Add comm-esr78 to the releases indexer

staktrace:


staktrace:
I didn't test this, so it might time out the indexer. But if so I'll just up the timeout again. I didn't think it was worth doing a test run (which takes a long time) just to see if it would timeout when we'll find out on the next releases run anyway.

PR #114, opened 2020-10-17
Bug 1670314 - Rename remaining config files.

staktrace:


staktrace:
I'm going to merge this now because it's trivial and needs to be done simultaneously with updating the lambda tasks, which I just did.

asutherland:
> I'm going to merge this now because it's trivial and needs to be done simultaneously with updating the lambda tasks, which I just did.

Agreed that this made sense.  And going further, I'm fine with post-landing review of stuff like this where:
- We expect any problems to be made obvious during the normal indexing runs that will succeed entirely or fail entirely without impacting indexing duration or qualitative indexing or UI changes.  Especially when...
  - In the event of failure, the worst-case is indexing data remains slightly stale, and we'll get email notifications.
  - Additional check runs or corrective runs are easily triggered.
  - The days that would be impacted by a problem are weekend days with massively reduced usage.

PR #115, opened 2020-10-19
Bug 1671917 - Add the mingw-w64 repo to searchfox

staktrace:


help.html
```diff
@@ -87,6 +87,7 @@ <h3>Repositories indexed</h3>
                                              version-control-tools</a></td> <td yes></td>   <td yes></td>   <td yes></td>   <td></td>       <td></td>       <td></td>       </tr>
        <tr><td><a href="/l10n/source/">l10n</a></td>                        <td yes></td>   <td></td>       <td yes></td>   <td></td>       <td></td>       <td></td>       </tr>
        <tr><td><a href="/llvm/source/">LLVM</a></td>                        <td yes></td>   <td yes></td>   <td></td>       <td></td>       <td></td>       <td></td>       </tr>
+       <tr><td><a href="/mingw/source/">MinGW-w64</a></td>                  <td yes></td>   <td yes></td>   <td></td>       <td></td>       <td></td>       <td></td>       </tr>
```
asutherland, line 90: mingw-moz should be added here too

mingw-moz/setup
```diff
@@ -0,0 +1,48 @@
...
+else
+    wget -nv https://s3-us-west-2.amazonaws.com/searchfox.repositories/mingw-blame.tar
+    tar xf mingw-blame.tar
+    rm mingw-blame.tar
+fi
+popd
+
+date
+
+MOZ_REV=$(curl -SsfL "https://hg.mozilla.org/mozilla-central/raw-file/tip/taskcluster/ci/fetch/toolchains.yml" | yq read - "mingw-w64.fetch.revision")
```
asutherland, line 37: I'm assuming there will be a separate pull request that adds `yq` to be installed in `indexer-provision.sh` (and the indexers re-provisioned) or in `indexer-update.sh` (and the indexers not re-provisioned).  (I probably should have explicitly made it clear that we don't install yq automatically yet.)

asutherland (approved):
This all seems reasonable, note 2 minor notes.

Thanks very much for proactively implementing all of these things and also entertaining my suggestions from the peanut gallery that double your work! ;)

help.html
staktrace: Ah, oops. Made the change and forgot to squash it into the force-push :) Done now

mingw-moz/setup
staktrace: Yeah, I filed another bug for that https://bugzilla.mozilla.org/show_bug.cgi?id=1672014 and it has a separate PR.

staktrace:
Had a small bug wherein I neglected to update the blame repo on the mingw-moz to the blame revision matching the MOZ_REV. That caused a panic because of a mismatch between the number of lines in the source file and the corresponding blame file, since the blame file it was looking at was of the wrong revision. Force-pushed a fix. Still waiting for a green test run.

PR #116, opened 2020-10-21
Bug 1672392 - Emit a warning instead of a hard failure if an l10n sub…

staktrace:
…repo disappears


staktrace:
This robustified against future failures of the same kind. I'll also re-upload the tarball with the x-testing subrepo removed.

asutherland (approved):

PR #118, opened 2020-10-28
Bug 958852 - Add the rust repo.

staktrace:
Requires https://github.com/mozsearch/mozsearch/pull/383

### DIFF
--- a/config4.json
+++ b/config4.json
@@ -40,6 +40,18 @@
       "objdir_path": "$WORKING/mingw-moz/objdir",
       "codesearch_path": "$WORKING/mingw-moz/livegrep.idx",
       "codesearch_port": 8084
+    },
+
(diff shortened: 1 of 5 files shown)


asutherland (approved):
Restating:
- We don't checkout the submodules because that would mean we lose blame info, accordingly, we tell codesearch / livegrep to not walk the submodules so that it doesn't fail.
- Even if we did support blame with submodules, given that https://bugzilla.mozilla.org/show_bug.cgi?id=958852#c7 indicates the rust repo's submodule includes a full checkout of LLVM, it's not really clear we'd want that to get fulltext indexed.  Or at least not without the ability to easily partition results such that LLVM is considered akin to test files or generated files.  That is, one wouldn't want LLVM results to overwhelm the rust results.

PR #119, opened 2020-11-11
Bug 1675511 - Copy some repos from config1.json to config4.json

staktrace:


asutherland (approved):
Thanks for laying out the plan in https://bugzilla.mozilla.org/show_bug.cgi?id=1675511#c1 r=asuth or rs=asuth for the corresponding removal step when it happens, etc.

PR #120, opened 2020-11-12
Bug 1675511 - Remove repos that were copied to config4 from config1

staktrace:


staktrace:
Carry rs=asuth from #119

PR #122, opened 2020-12-04
Fix process-tc-artifacts.sh for rust 1.47 stdlib.

emilio:
In rust 1.47 the library is indeed in /library, but there's also
stuff under /src (just `llvm-project/libunwind`).

Since it's not empty we were copying it instead of the rust stdlib.

We could copy both instead, though that might need adding wildcards and
complicating the script. This fixes the indexing error and we can do
that in a follow-up, though it's not clear that having libunwind gives
us much benefit.


staktrace:
Was there anything in `/library` prior to 1.47? This code runs on beta/release trees too so on older trees it may copy the wrong thing if there's stuff in `library` there. Otherwise r+

emilio:
Took a look at the esr68 indexer, nothing there.

emilio:
There as in `/library` :)

PR #123, opened 2020-12-28
Bug 1684371 - Update yq syntax for yq version 4.2.0

staktrace:
I tested by running the command in question on today's release2 indexer which is still going


asutherland (approved):
Thanks for the fix!

staktrace:
I'll merge this now while I figure out how to not use the snap.

PR #124, opened 2020-12-28
Bug 1683563 - Add a script that facilitates rebuilding gecko-blame

staktrace:
This script currently isn't invoked from anywhere but it can be run
manually to regenerate the gecko-blame repo. Future changes might
make this easier to run in an unattended fashion.


staktrace:
rs=asuth from https://bugzilla.mozilla.org/show_bug.cgi?id=1683563#c9

PR #125, opened 2020-12-28
Bug 1684386 - Pack the kaios tarballs more

staktrace:
This copies over the gc heuristics from the mozilla-central upload
script to kaios as well, since it suffers from the same problem
where the tarballs can grow quite large over time.

### DIFF
--- a/kaios/upload
+++ b/kaios/upload
@@ -9,6 +9,18 @@ date
 echo Uploading kaios
 pushd $INDEX_ROOT
 tar cf kaios.tar git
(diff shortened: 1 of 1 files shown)


staktrace:
Haven't tested this yet.

staktrace:
I kicked off a release4/config4 indexing run using this branch to verify it works. The upload scripts don't run on non-release channels so it was easier to just use a release channel.

staktrace:
Verified this works.

asutherland (approved):

PR #127, opened 2021-01-29
Bug 1689747 - Update llvm to main branch

staktrace:


staktrace:
I'm going to go ahead and merge this so it's in for tomorrow's run. Yell at me if I screwed up!

asutherland:
Thank you for the fix and terminating the indexer!  I saw the email and presumed this was what had happened but had not gotten to the first investigatory steps (which would then have been followed by sending a lock-acquiring email like you sent ;)

PR #128, opened 2021-02-02
Bug 1690181 - Normalize the config4.json repos git/blame folders

staktrace:
This makes it so that all the repos listed in config4.json extract
their git tarballs into a `git` folder and their blame tarballs into
a `blame` folder. This is good for consistency and scriptability.
Also the tarball names are now `<repo-name>.tar` for the git tarball
and `<repo-name-blame>.tar` for the blame tarball.

### DIFF
--- a/config4.json
+++ b/config4.json
@@ -56,8 +56,8 @@
 
     "whatwg-html": {
       "index_path": "$WORKING/whatwg-html",
-      "files_path": "$WORKING/whatwg-html/html",
-      "git_path": "$WORKING/whatwg-html/html",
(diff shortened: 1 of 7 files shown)


staktrace:
I'll need to manually download/rename/reupload the tarballs concurrently with merging this, so I'm doing it one config file at a time.

staktrace:
Also that's why I'm starting with config4.json, so I can work out kinks in the process with the low-value repos first :)

asutherland (approved):
Woo!  I eagerly look forward to the inevitable benefits from these normalizations!

PR #129, opened 2021-02-05
Bug 1690181 - Rename gecko-dev.tar to gecko.tar

staktrace:
All the other repos have their git repo in `<foo>.tar` and
blame repo in `<foo>-blame.tar`. gecko-dev is unique in using
`gecko-dev.tar` and `gecko-blame.tar`. This patch brings gecko
in line with the other repos, by changing the git tarball to
be `gecko.tar`. Note that the folder inside is still `gecko-dev`;
eventually it will be renamed to `git` to match other
repositories but that'll happen when I'm more in the mood to
manually download and re-upload a 10G tarball.

PR #130, opened 2021-02-07
Bug 1683563 - Add blame rebuilder scripts for all repos

staktrace:
This generalizes the mozilla-central/rebuild-blame.sh script into a
shared/rebuild-blame.sh script that can be used to do a blame rebuild
for any repo, assuming certain conventions (which hold for all of our
repos).

Each repo gets a "reblame" script that invokes the main script with
the necessary options.

A couple of things to note:
- The intent is that when making changes to the build-blame tool that
  affect historical blame (e.g. like changing copy detection thresholds
  or such), we will be able to trigger each repo's reblame script with
  the --upload --in-place arguments, which will rebuild the blame repo
  and upload it to S3.
- In some cases, we may want to make changes to the blame repo that
  are not backwards compatible or that require additional testing. For
  these cases, the reblame script can be used with just --upload, which
  will upload the blame tarballs so as to not clobber the existing ones.
  Testing can then be done as needed, and/or the new tarballs can be
  rotated into place simultaneously with changes to the code that
  consumes the blame information. This should help avoid the need for
  the consumer code to deal with both the "old" and the "new" blame
  formats.


staktrace:
Just FYI this PR stacks on top of #129 (I pushed the commits from that PR into a "normalize" branch on this repo, and used that branch as the base branch for this PR, but after merging #129 we should change the base branch on this PR back to master)

asutherland (approved):
This looks great and is very exciting progress!  Thank you!

PR #131, opened 2021-02-10
Bug 1691724 - Failover to using "latest" for bugzilla component mappings when the indexed revision artifact isn't there. r=kats

asutherland:
I also updated the coverage comment which was a little bit-rotted.


staktrace (approved):
LGTM, thanks!

PR #132, opened 2021-02-11
Bug 1691239 - Run git gc prior to repo uploads on saturday AM runs.

staktrace:


staktrace:
This still needs some testing, but putting it up for now.

asutherland (approved):
Looks good!

Maybe we should add some date-printing commands to git-maintenance.sh so we can eyeball how long the gc's are taking? Also fine if not, as I think we currently don't care and if we started caring about such things the right course of action would probably be some more aggressive tool that does cool timeline charts like https://www.bootchart.org/ (but which I guess is an artifact of the past at this point?).

staktrace:
I threw a couple of `date` commands in there, good call on that. It won't hurt and gecko-dev/kaios might have significant gc times so it would be worth at least knowing about that.

PR #133, opened 2021-02-12
Normalize folder names in comm-central and mozilla-mobile

staktrace:
Merging these patches will need to be accompanied with actual changes to the tarballs, which I will do once the patches are r+'d.


asutherland (approved):
Thanks!  Also this is a great TGIF landing! ;)

PR #135, opened 2021-02-20
Normalize gecko.tar and gecko-blame.tar

staktrace:
This needs to merge concurrently with changes to the tarballs. I tested it using kats-channel indexer runs for config1.json and config2.json.


asutherland (approved):
Thanks!  I think there will be a release1 indexer starting now-ish per the now up-to-date https://github.com/mozsearch/mozsearch-mozilla#how-searchfoxorg-stays-up-to-date

PR #136, opened 2021-02-26
Bug 1683563 - Minor tweaks to the rebuild-blame script

staktrace:
- Silence curl's logging spew
- Do an aggressive gc after building the blame repo, which
  in particular produces better delta compression and can result in
  a significantly smaller repo


asutherland (approved):
Gc Aggressive! G C Aggressive![1]

1: I think this is a bring it on reference

PR #137, opened 2021-05-12
Bug 1690181 - Follow-up to properly exclude the mozilla subfolder fro…

staktrace:
…m the comm-central tarball


staktrace:
I noticed the comm-central.tar tarball in S3 is > 10G big, and realized it was because we're not excluding the mozilla/ subdir. Whoops!

emilio (approved):

emilio:
Thank you!

PR #138, opened 2021-05-22
Bug 1683563 - Don't use aggressive GC since it can OOM

staktrace:


asutherland (approved):
This is of course the right technical and procedural fix, but every time I saw this my brain went to the cheer-leading cheers of "Be Aggressive, B E Aggressive!" but with s/be/gc/ and so I will miss this foot-gun of a setting.  (I'm not sure if I ever managed to work that into a review comment or not against my better judgement.)

asutherland:
I totally did make the reference!  https://github.com/mozsearch/mozsearch-mozilla/pull/136

PR #139, opened 2021-05-24
Bug 1712622 - Index the m-c pine project branch. r=kats

asutherland:
I know for "ash" we just called it "ash", but it seems like having the
"mozilla-" prefix better conveys that's it's an m-c branch and that it
uses the same underlying gecko repo/etc. while avoiding the problems
of using "mozilla-central" as a prefix like in "mozilla-central-pine".

I put this in config4 because config4 currently has the fastest indexing
time per my analysis at
https://bugzilla.mozilla.org/show_bug.cgi?id=1712622#c1 and our disk
usage is also relatively low, at 41G / 240G

I'm intentionally leaving this off the help.html listing for now per
discussion with the branch owners.

### DIFF
--- a/mozilla-central/reblame
+++ b/mozilla-central/reblame
@@ -9,5 +9,5 @@ set -o pipefail # Check all commands in a pipeline
 # as to minimize the amount of new work needed. The order of branches
 # in the loop is also selected to reduce unnecessary work; changing the
 # order should not affect correctness but may increase redundant work.
-BRANCHES="beta release esr78 esr68 esr60 esr45 esr31 esr17"
+BRANCHES="pine beta release esr78 esr68 esr60 esr45 esr31 esr17"
 $CONFIG_REPO/shared/rebuild-blame.sh --tarball-base gecko --branches "${BRANCHES}" $*
--- a/mozilla-pine/setup
+++ b/mozilla-pine/setup
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+# See ../mozilla-central/setup for documentation
+# The mozilla-central repo also uploads an updated gecko.tar which
+# includes pine commits, so we don't need our own upload script.
+
+REVISION_TREE=pine
+REVISION_ID=latest
+
+date
+
+source $CONFIG_REPO/shared/resolve-gecko-revs.sh $REVISION_TREE $REVISION_ID
+
+date
+
+$CONFIG_REPO/shared/checkout-gecko-repos.sh $REVISION_TREE "pine" "$INDEXED_HG_REV"
+
+date
+
+$CONFIG_REPO/shared/fetch-tc-artifacts.sh $REVISION_TREE $INDEXED_HG_REV "$PREEXISTING_HG_REV"
+
+date
--- a/config4.json
+++ b/config4.json
@@ -118,6 +118,17 @@
       "objdir_path": "$WORKING/kaios/objdir",
       "codesearch_path": "$WORKING/kaios/livegrep.idx",
       "codesearch_port": 8091
+    },
+
+    "mozilla-pine": {
+      "index_path": "$WORKING/mozilla-pine",
+      "files_path": "$WORKING/mozilla-pine/git",
+      "git_path": "$WORKING/mozilla-pine/git",
+      "git_blame_path": "$WORKING/mozilla-pine/blame",
+      "hg_root": "https://hg.mozilla.org/projects/pine",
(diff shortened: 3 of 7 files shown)


asutherland:
I've done a try triggering of config4 on the "asuth" channel with the understanding that between the lack of an "upload" script and this not being the release channel, no uploads should happen and mess anything up.

Based on our approach with ash and the general blame overlap I went with treating this similarly to that other than the naming scheme change I mention above.

In general my approach was to copy what we did with esr17 and then remove the gecko-dev link from the config file because the project branches don't appear to be replicated into gecko-dev.

staktrace:
Your test run failed with this:

```
++ curl -SsfL --compressed https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.pine.latest.firefox.linux64-searchfox-debug/artifacts/public/build/target.json
curl: (22) The requested URL returned error: 404
```

So I guess either we need to turn off some analysis platforms for pine, or we need to enable the searchfox taskcluster jobs on that branch.

mozilla-pine/setup
```diff
@@ -0,0 +1,26 @@
...
+
+source $CONFIG_REPO/shared/resolve-gecko-revs.sh $REVISION_TREE $REVISION_ID
+
+date
+
+$CONFIG_REPO/shared/checkout-gecko-repos.sh $REVISION_TREE "pine" "$INDEXED_HG_REV"
+
+date
+
+$CONFIG_REPO/shared/fetch-tc-artifacts.sh $REVISION_TREE $INDEXED_HG_REV "$PREEXISTING_HG_REV"
```
staktrace, line 24: I think you might need to drop this call, unless you plan on enabling all the necessary jobs on the pine branch? Not sure.

mozilla-central/reblame
```diff
@@ -9,5 +9,5 @@ set -o pipefail # Check all commands in a pipeline
 # as to minimize the amount of new work needed. The order of branches
 # in the loop is also selected to reduce unnecessary work; changing the
 # order should not affect correctness but may increase redundant work.
-BRANCHES="beta release esr78 esr68 esr60 esr45 esr31 esr17"
+BRANCHES="beta release esr78 esr68 esr60 esr45 esr31 esr17 pine"
```
staktrace, line 12: I would put `pine` at the start of this list since it's probably pretty close to m-c. Putting it at the end means the code will start with the esr17 blame branch and build up pine from there, which will be a lot more work.

asutherland:
I'll enable the searchfox builds on the branch since I probably would end up needing to do that anyway.

Thank you for the high quality review, especially the blame ordering nuance which I hadn't grasped.

mozilla-pine/setup
asutherland: I enabled searchfox jobs on pine and it seems to have worked, with results showing up on https://treeherder.mozilla.org/jobs?repo=pine today.

asutherland:
@staktrace Okay, finally got the indexing to pass.  I added 2 hack commits in here I won't land (placing it in config1 as well so the pine blame can be used, and then stealing the blame directory from mozilla-central into mozilla-pine), but https://github.com/mozsearch/mozsearch-mozilla/pull/139/commits/cdd530fec5be64783977b21cf258f10136d05e1d is new and maybe useful?  Specifically, letting it create the blame branch automatically without me having to manually mess with the tarball.

staktrace (approved):
Yep, this looks good! I forgot that new branches off m-c still required manual work (it's been so long since I had to do that) and this is a nice fix to that.

asutherland:
I landed this now, so this will take effect in ~20 minutes when the config1 indexer runs ([docs on this](https://github.com/mozsearch/mozsearch-mozilla#how-searchfoxorg-stays-up-to-date) continue to be up-to-date).  I'll manually re-trigger the config4 indexer once config1 is passed the upload stage.  (config4 is basically already done for today's run, so it will serve as a timely backup if I did mess things up somehow).

This does make me realize that the "config1 updates everyone's blame for them ahead of time" is slightly now wrong because config1's "morning" run (UTC10) starts after every other morning run now because it waits for blame data, but I think this is likely fine because:
- it does still process blame for the UTC22 build at UTC4, which should generally minimize drift.
- mozilla-central is really the only branch with any high levels of expected turnover.  beta will mainly see uplifts, and I understand the current set of people who would land these to be spread mainly between the US (not a timing problem) and Europe (yes, some double-blame-building will occur).  Still, this is like O(10) per day.

asutherland:
As a note for posterity: One should not merge changes to a config file after an indexer for it has started and before it has spawned its web-server, as the indexer will hang while waiting for the web-server which will have a bad day when being told about a tree that does not exist in the index.

I've terminated the hung config4 indexer and web-server and re-triggered config4 now that the upload finished.  I'm also going to do the ELB changes now.

PR #140, opened 2021-05-27
Bug 1683563 - Flip-flop back to a git repack

staktrace:
The aggressive gc we were using before was useful in significantly
reducing the size of the blame tarballs. However, we can achieve
the same effect with an explicit repack, where we control the memory
usage better to avoid OOMs. See comments in the code for details.

### DIFF
--- a/shared/rebuild-blame.sh
+++ b/shared/rebuild-blame.sh
@@ -116,7 +116,21 @@ for BRANCH in $BRANCHES; do
 done
 
 pushd "${BLAME_REPO_DIR}"
-git gc
+# After building the blame with fast-import the pack delta compression is
+# generally quite bad. We can reduce the space requirement by asking
+# git to repack the repository and compute new delta compression chains.
+# See https://stackoverflow.com/a/28721047 for a more detailed discussion.
+# This will often reduce space requirement by 2x or more. However, it has
+# a tendency to OOM, so we restrict the amount of memory allowed to 3G
+# per thread. On our current indexer instances (8 cores, 32G memory) this
+# works well. It still causes an OOM on the whatwg-html blame repo, because
+# it happens while writing the repacked objects rather than during
+# the repacking itself. So what we do is first try with 3G per thread, and
+# if that fails, try again with 2G per thread. This gives us the benefit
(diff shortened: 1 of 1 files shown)


shared/rebuild-blame.sh
```diff
@@ -116,7 +116,21 @@ for BRANCH in $BRANCHES; do
...
-git gc
+# After building the blame with fast-import the pack delta compression is
+# generally quite bad. We can reduce the space requirement by asking
+# git to repack the repository and compute new delta compression chains.
+# See https://stackoverflow.com/a/28721047 for a more detailed discussion.
+# This will often reduce space requirement by 2x or more. However, it has
+# a tendency to OOM, so we restrict the amount of memory allowed to 3G
+# per thread. On our current indexer instances (8 cores, 32G memory) this
+# works well. It still causes an OOM on the whatwg-html blame repo, because
+# because it happens while writing the repacked objects rather than during
```
asutherland, line 127: nit: there's two becauses in a row here, separated by the line wrap

shared/rebuild-blame.sh
```diff
@@ -116,7 +116,21 @@ for BRANCH in $BRANCHES; do
...
+# the repacking itself. So what we do is first try with 3G per thread, and
+# if that fails, try again with 2G per thread. This gives us the benefit
+# of faster/better repacking on repos that don't OOM with 3G, and falls
+# back to 2G if that fails. If we find repos that OOM even with 2G we can
+# reduce that further on the fallback call.
+git repack -f -a -d --depth=250 --window=250 --window-memory=3g || git repack -f -a -d --depth=250 --window=250 --window-memory-2g
```
asutherland, line 133: I think this is a typo in the failover? `--window-memory-2g` (there's a dash before 2g instead of an equal?)

asutherland (approved):
Thanks for the detailed comment and thanks VERY much for all your work on the blame rebuilding!

shared/rebuild-blame.sh
staktrace: Nice catch, on both errors!

PR #145, opened 2021-07-29
Bug 1722889 - Update check target to one that still exists

staktrace:
In bug 1711648 the Box2d type was removed from the webrender
bindings code, so this check started failing. Updating it to
reference a different type from the same file that is less
likely to be deleted.


staktrace:
Merging since it's a pretty trivial fix

PR #146, opened 2021-08-25
Bug 1707282 - convert mozilla-central checks to insta

asutherland:
This includes the result of running the following command to create the
snapshots:

```
./review-build-check-results.sh config1.json mozilla-central release
```

One "oh yeah, that's a thing" surprise was that mozilla-central
includes coverage data in the HTML output, so I modified the test
mechanism to strip the coverage and blame lines.


asutherland:
This passed a try push once all the rustc nightly issues went away.  I manually verified that my run ran the `test_check_glob` for the local indexer case (which was visible in the `index-log`) as well as the web-server check which I had to look at the logs uploaded to S3 to see because the output of trigger-web-server.py doesn't get reflected into the index-log out of necessity.

PR #147, opened 2021-08-31
Bug 1726109 - Add comm-esr91 to config2

staktrace:


asutherland (approved):
Thank you!  I wasn't quite sure how to do this but was hoping the trigger_shell thing would make it easier to noodle around and figure out, and in a way, it did, because you did it!

I share your concerns about the runtime of config2, but do think it's probably still the best slot for now given that the mozilla-pine repo in config4 is relatively important and it's desirable for that to have a relatively low latency.

If we start hitting the 10 hour limit I can certainly bump the number at https://github.com/mozsearch/mozsearch/blob/master/infrastructure/aws/trigger_indexer.py#L10 and regenerate the lambda job for that indexer.

staktrace:
I had almost forgotten how to do this too, but it's actually reasonably documented :)

asutherland:
> I had almost forgotten how to do this too, but it's actually reasonably documented :)

Ah, yeah, I should clarify that I was not concerned about the level of documentation or clarity of implementation on the searchfox side of the equation as much as I was concerned that I had vague memories of people talking about overhauling the relationship between the comm-central tree and the mozilla tree and I was worried the new esr might be different from the previous esr's for that reason.

PR #148, opened 2021-09-02
Bug 1716167 - Run maintenance for submodules as well

staktrace:
For l10n, the submodules are cinnabar repos and need
regular cinnabar fsck runs to avoid warnings, so let's do
that.

Even ignoring the cinnabar warnings, doing regular gc on
the submodules is a good idea, so we do it for mozilla-mobile
as well. In that case there's no cinnabar submodules so
we can use the more normal foreach subcommand rather than
manually looping through the submodules.

### DIFF
--- a/l10n/upload
+++ b/l10n/upload
@@ -8,10 +8,19 @@ date
 
 echo Uploading l10n
 pushd $INDEX_ROOT
+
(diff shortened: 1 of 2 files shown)


asutherland (approved):
Thanks!

PR #174, opened 2022-07-24
Bug 1759071 - Use coverage aggregation jobs as basis for m-c rev selection.

asutherland:


asutherland:
I'm going to do a test run immediately-ish with some mozsearch changes I'm also landing.

PR #184, opened 2022-09-25
Bug 1792322 - Actually, we can just use `+` in sed because of `-E`

asutherland:
I had already dealt with `+` systematically by having us use "-E" so my attempt to make sure sed was happy with `\+` actually made things worse, whoops!


asutherland:
I tested this locally this time.  In general this is probably an argument that our mozilla-central machinery should probably be moved into `mozsearch` with explicit but minimal test coverage, somewhat like how we do with all the test and coverage machinery that derive-per-file-info ingests.

PR #194, opened 2023-02-21
Check coverage revision buildid to ensure it's recent enough

emilio:


asutherland:
Very nice failsafe fix!  I'd never seen or expected anything as dramatic as this most recent instance!  In general it seemed like the flakiness of the coverage aggregation was at worst a few days, but from https://chat.mozilla.org/#/room/#taskcluster:mozilla.org it sounds like something very weird happened.  In general it was my understanding that coverage would run when all of the jobs for a treeherder push had completed, which, if a job didn't successfully complete before its timeout 24 hours later, could result in aggregation running after a 24 hour delay and causing (bounded!) time travel.  It sounds like the expiration rules maybe allowed for a much more extreme version of time travel and this seems like a very appropriate and clearly necessary mitigation given that there have been no indications of coverage stability improving.

One thing worth considering might also be building off the script as I proposed at https://chat.mozilla.org/#/room/#taskcluster:mozilla.org which could let searchfox be less confused in situations like this where code coverage runs out of order and makes the "latest" task useless.  It's also potentially important to note that the code coverage sets "latest" when it starts the job, I think, not when it completes.

PR #196, opened 2023-03-29
Stop trying to build glean

staktrace:
This isn't working now anyway, because the nightly rust compiler has dropped support for save-analysis. And without that there's no point trying to compile.


staktrace:
Going to merge since it's pretty trivial.

PR #200, opened 2023-06-12
Turn of JXL in webkit build.

emilio:
As we don't have libjxl on the bots. Maybe it's a matter of updating the flatpak?


asutherland:
@emilio We removed the flatpak stuff in https://github.com/mozsearch/mozsearch-mozilla/pull/180 where @dpino provided some amazing context.

I think as explained there (AIUI), we mainly expect to get our dependencies from [our invocation of Tools/gtk/install-dependencies](https://github.com/mozsearch/mozsearch-mozilla/blob/8c5ca8f290508b364e1b71b129ccdabc43a97916/wubkat/setup#L82) which [does some stuff](https://searchfox.org/wubkat/source/Tools/gtk/install-dependencies) to [install this list of "gtk" packages](https://searchfox.org/wubkat/source/Tools/gtk/dependencies/apt) which also depends on [this list of "glib" packages](https://searchfox.org/wubkat/source/Tools/glib/dependencies/apt).

Maybe it would make sense for someone more involved in the webkit community to be on the searchfox-aws list like @dpino so they can get copies of the build failures?  I think before webkit-search.igalia.com started redirecting to searchfox.org it was more obvious when that package list would need to be updated.  This might also be a stepping stone to adding someone webkit focused to the people.mozilla.org searchfox group so they could do dev runs of searchfox to try out changes to the wubkat scripts and/or create a wubkat-specific [per-file-info.toml](https://github.com/mozsearch/mozsearch/blob/master/config_defaults/per-file-info.toml) which will become more relevant if any of my local experiments end up making it upstream like [teaching searchfox how to understand runnables](https://mozilla.social/@asutherland/110189570586792775) which are inherently tree-specific.  It would be mutually beneficial for everyone if the webkit community could be more involved in wubkat.

dpino:
@asutherland @emilio Right, we eventually decided to redirect https://webkit-search.igalia.com to https://searchfox.org/wubkat/source since we could not maintain the site with reasonable uptimes.

Definitely I'd be interested in getting subscribed to searchfox-aws mailing list (how can I do that?). I'm the person in charge at Igalia of maintaining the Ubuntu bots, so I'm constantly keeping an eye on them. The other day when the Ubuntu 22.04 upstream bots failed after https://github.com/WebKit/WebKit/pull/14790 landed, I fixed the Ubuntu 22.04 bots but it completely slipped my mind that change was also affecting webkit-search. Alternatively, if you notice webkit-search has failed but [[GTK-Linux-64-bit-Release-Ubuntu-LTS-Build](https://build.webkit.org/#/builders/68)] is green, please feel free to ping me on WebKit's Slack. It's likely I have changed something inside the bot.

FWIW, the change look goods to me. I implemented the same change in the upstream bots. Since there's no libjpegxl package in Ubuntu 22.04 the fix is to pass a build flag to disable JPEGXL.

PR #208, opened 2023-09-05
Bug 1851527 - Index cedar

glandium:


shared/collapse-generated-files.sh
```diff
@@ -74,8 +74,11 @@ fi
 
 # The generated file was not the same across all platforms, so
 # put the different versions in __$PLATFORM__ subfolders.
-for PLATFORM in linux64 macosx64 win64 android-armv7; do
+for PLATFORM in linux64 macosx64 win64 android-armv7 ios; do
     if [ ! -f "generated-$PLATFORM/$GENERATED_FILE" ]; then
+	if [ $PLATFORM = ios ]; then
```
emilio, line 79: Fix indentation?

emilio (approved):
Looks sensible. The iOS changes could've been their own PR but sure.

asutherland:
Is there someone we should add to https://groups.google.com/a/mozilla.com/g/searchfox-aws where the automatic warning/error emails get sent so they can see if warnings/errors are reported for cedar?

shared/collapse-generated-files.sh
asutherland, line 79: This change seems like it merits a comment explaining what's going on.

asutherland:
tl;dr: config-choice wise, config4 seems good.

I just did a quick run of `scripts/indexer-logs-fetch.sh` to get the most recent config4 log and it seems like the first `date` to last `date` delta is currently ~4h17m for config4 which seems fine (which makes sense because that's where pine was removed from).  config2 is currently a little bit faster at ~3h52m, but config2 also has mozilla-beta and mozilla-release which are more important to not have potentially break.

shared/collapse-generated-files.sh
glandium: Note that practically speaking, it's not useful because ios is last, but in case it stops being last...

asutherland:
> Is there someone we should add to https://groups.google.com/a/mozilla.com/g/searchfox-aws where the automatic warning/error emails get sent so they can see if warnings/errors are reported for cedar?

I added @glandium to the group and invited him to the searchfox-aws LDAP group and provided the ssh key via slack[1].

1: I think I've noted elsewhere that we can probably stop using that key in favor of just having the ssh script add a public key signature to the EC2 instance on demand for the user.  If we're concerned about the privacy signature for that, the ssh script could create its own private key on demand on first use or something like that.

PR #224, opened 2024-02-06
Bug 1878998 - Add comm-beta and comm-release to Searchfox indexing

jfx2006:
Thunderbird is moving to a monthly release schedule the same as Firefox's. Part of this has been to revive the comm-release repository.
This patch adds comm-beta and comm-release to Searchfox in the same manner as comm-esr115 (no mozilla/ subtree).

### DIFF
--- a/comm-beta/build
+++ b/comm-beta/build
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
--- a/comm-beta/reblame
+++ b/comm-beta/reblame
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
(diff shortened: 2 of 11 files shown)


asutherland:
Please file a bugzilla bug for tracking and discussion purposes; https://bugzilla.mozilla.org/show_bug.cgi?id=1836844 is the example from comm-esr115.  That said, this should be fine, the bare comm repositories are quite small (1.6G for comm-esr115 with a ligvegrep.idx of 520M), and we can move both mozilla-esr102 and comm-esr102 to config3.

config3 is currently at 187G used of 295G, 93G available.  comm-esr102 is 1.5G so that's fine.  mozilla-esr102 is at 33G which is also fine, but probably the last thing we can move in there before doing something about storage.  I think https://bugzilla.mozilla.org/show_bug.cgi?id=1517978 will likely force the issue and we'll need to be more clever for the gecko "git" and "blame" subdirectories which are naively replicated (but with differing checkouts for "git").  Options there are mainly [worktrees](https://git-scm.com/docs/git-worktree) and just specifying the git/blame (well, "history" after that bug) paths as a single unified path that's not under INDEX_ROOT.

asutherland:
going to close this for now since we have https://bugzilla.mozilla.org/show_bug.cgi?id=1878998 tracking this

PR #226, opened 2024-02-10
graphviz: Remove submodule steps

Smattr:
The Graphviz submodules are only relevant to users building on Windows platforms.

### DIFF
--- a/graphviz/setup
+++ b/graphviz/setup
@@ -37,8 +37,6 @@ date
 echo Updating git
 pushd $GIT_ROOT
 git pull origin main
-# graphviz uses submodules
-git submodule update --init --recursive
 
 popd
 


asutherland:
Thanks!  I think I'll need to manually clean up the git repo (we propagate checkouts forward via tarball) if we land this.

But before landing: @Smattr one potentially nice aspect of having the submodules checked out and up-to-date is that even though searchfox's blame processing doesn't understand submodules and we don't have a windows build run here, the fulltext search index built using livegrep's codesearch backend still works.  So like [searching for FLEX_STD](https://searchfox.org/graphviz/search?q=FLEX_STD&path=&case=false&regexp=false) will find [windows/dependencies/graphviz-build-utilities/FlexLexer.h](https://searchfox.org/graphviz/rev/85071508d642cfe7cdecaa2c2291acbc494347f4/windows/dependencies/graphviz-build-utilities/FlexLexer.h) which might be useful?  Is it definitely not useful to have the contents of these submodules around?  I should note that searchfox has a concept of "path kinds" where we could mark these directories as third-party so that they come much lower in the search results ranking.

For example, for mozilla-central [searching for sqlite3_init](https://searchfox.org/mozilla-central/search?q=sqlite3_init&path=) well, finds our NSS copy of sqlite is miscategorized, but if collapsed/scrolled past, our canonical sqlite3 vendoring is in the third-party section.

Smattr:
I guess this is subjective. Some headers are there, but the actual implementations of many functions will be unavailable, as they’re only present in the repo compiled into binary libs. These are also only Windows headers. For some dependencies, it’s possible they ship different headers on other platforms. I leave it up to you to decide whether to retain these or not

Smattr:
As an anecdotal aside, having mismatched Flex headers in the compiler’s include path during compilation has actually been a significant headache for us in the past. See point (3) in [commit b64593afc07d3f62a13a3b6b634bcb80b7fe6180](https://gitlab.com/graphviz/graphviz/-/commit/b64593afc07d3f62a13a3b6b634bcb80b7fe6180).

asutherland:
Thanks for the context!  It definitely sounds like having the submodules present hurts more than it helps given that mozsearch is only running a linux build/indexing pass.

I am going to try and coordinate the landing and related tarball changes in a batch pass with some other tree management stuff in the next week or two.

tutjaaye (approved):

asutherland:
Okay, I regenerated the tarball from a fresh clone without having done any submodule checkouts, so merging this now.

asutherland:
An unforeseen side-effect of the change was that the fulltext codesearch was not happy because we tell it to walk all submodules by default and because the webkit tree on the server we use for graphviz is frequently flaky, I did not notice immediately; I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1904890 and there are fixes up there that should land soon and ideally take effect tomorrow.

PR #230, opened 2024-03-19
Revert "Avoid warnings about missing ios on non-cedar branches"

glandium:
This reverts commit d8202392c2847c1e01d59cb34a5608f621312f85.

Now that mozilla-central has ios searchfox tasks, we shouldn't need to exclude those.


asutherland:
As discussed on slack, we do expect this to produce warnings on branches where we don't have ios builds, but we already have a bunch of warnings and presumably we'll end up with ios builds everywhere eventually.

PR #232, opened 2024-05-06
Bug 1895316 - Fetch chrome-map.json

arai-a:
Preparation part for https://bugzilla.mozilla.org/show_bug.cgi?id=1895316

This downloads `chrome-map.json` from ccov task.
the processing part is addressed in mozsearch repository.


asutherland (approved):
Sorry, forgot about this one!

PR #235, opened 2024-05-15
Exclude WPT from HTML indexing

arai-a:
Given there are so many HTML files in WPT, and the main purpose here for indexing HTML is to linkify chrome/resource URLs, which won't appear in WPT files, I'll exclude them for now.
We can revisit this once the crossref gets chunked and parallelized, and made capable of handling more files with less RAM usage

### DIFF
--- a/comm-central/repo_files.py
+++ b/comm-central/repo_files.py
@@ -17,3 +17,8 @@ def filter_js(path):
     if 'js/src/tests' in path or 'jit-test' in path:
         return False
     return True
+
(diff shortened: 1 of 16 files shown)


asutherland (approved):
Reviewer note: https://github.com/mozsearch/mozsearch/pull/722 introduces the ability to use `filter_html`.

I agree there's no upside to trying to index the WPT HTMLs given our current indexing functionality and limitations.

PR #236, opened 2024-05-17
Bug 1795177 - Fetch and process macos64-aarch64 artifacts.

arai-a:
Depends on https://phabricator.services.mozilla.com/D210841


asutherland (approved):
Thanks so much!

PR #238, opened 2024-05-23
Exclude WPT from CSS indexing

arai-a:
Same as HTML, indexing CSS inside WPT won't add much value right now.

required by https://github.com/mozsearch/mozsearch/pull/733

### DIFF
--- a/comm-central/repo_files.py
+++ b/comm-central/repo_files.py
@@ -22,3 +22,8 @@ def filter_html(path):
     if 'testing/web-platform/' in path:
         return False
     return True
+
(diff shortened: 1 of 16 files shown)


emilio (approved):

asutherland:
I'm going to land this now so I can trigger a dev channel run of https://github.com/mozsearch/mozsearch/pull/733.  (`trigger_indexer.py` and the scripts it propagates information to only can handle a single branch name, so when the branch names diverge like in this case it doesn't work.)

PR #242, opened 2024-05-29
Bug 1899642 - Use ld-wrapper, and fix git URL

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1899642
depends on https://github.com/mozsearch/mozsearch/pull/741

This does:
  * use ld-wrapper added by https://github.com/mozsearch/mozsearch/pull/741
  * fix the git URL, which had been causing the redirection warning


asutherland (approved):
Thank you so much!

PR #245, opened 2024-06-09
Bug 1416899 - Add filter for WebIDL

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1416899

This does the following:
  * add filter for webidl files, and filter out test files and also unused files
  * add generated `CSS2Properties.webidl` to the list, in recent tree


asutherland (approved):
The prospect of having searchfox understand WebIDL is very exciting!

PR #246, opened 2024-06-14
Bug 1901722 - Move esr102 repos to config3

jcristau:
They're not getting updated any more, so they don't need to be rebuilt every day.


asutherland:
Closing this since I merged them all in https://github.com/mozsearch/mozsearch-mozilla/pull/248 because of the limitations of github's support for stacks.

PR #247, opened 2024-06-14
Bug 1901722 - Start pulling esr128

jcristau:
Depends on #246.

PR #248, opened 2024-06-14
Bug 1901722 - Add esr128 repos

jcristau:
Depends on #247.


asutherland:
Thanks for the changes!  For simplicity I'm going to merge this and then do the quick lz4 usage fixup since I ended up using lz4 for the new tarballs as part of a gradual improvement thing.  (This improves s3 download/upload times and s3 storage use.)

PR #253, opened 2024-06-23
Bug 1416899 - Update WebIDL filters to handle tools/ts, missed dirs

asutherland:
This changes the m-c repo_files to ignore tools/ts added by https://bugzilla.mozilla.org/show_bug.cgi?id=1895418 which caused breakage.  This also propagates the updated repo_files to the other actively indexed m-c trees which includes propagation of the CSS2 properties which seems to be on all branches so it's okay to add.

### DIFF
--- a/mozilla-beta/repo_files.py
+++ b/mozilla-beta/repo_files.py
@@ -10,6 +10,8 @@ def filter_webidl(path):
         return False
     if 'dom/webidl/MozApplicationEvent.webidl' in path:
         return False
+    if 'tools/ts/' in path:
+        return False
     return True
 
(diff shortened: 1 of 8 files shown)


asutherland:
Ah, I missed filtering it out for comm-central but the failure was downgraded to a warning by tree rules.  (Only comm-central has the "mozilla" check-out for the comm-* trees.)

asutherland:
Addressed in https://github.com/mozsearch/mozsearch-mozilla/pull/254 hopefully

arai-a:
Thank you for taking care of this!

PR #257, opened 2024-07-08
Bug 1858453 - Index mozsearch and mozsearch-mozilla

asutherland:
Try run for config1.json with trees is up on asuth.searchfox.org :
- https://asuth.searchfox.org/mozsearch/source/
- https://asuth.searchfox.org/mozsearch-tests/source/
- https://asuth.searchfox.org/mozsearch-mozilla/source/

### DIFF
--- a/config1.json
+++ b/config1.json
@@ -74,7 +74,7 @@
     },
 
     "mozsearch": {
-      "priority": 500,
+      "priority": 600,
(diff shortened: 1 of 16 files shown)


asutherland:
I terminated the just-started config1 indexer and re-triggered so it gets this change so I can be around if there are any hiccups.

PR #260, opened 2024-07-23
Bug 1883953 - Add android-aarch64.

nicolas-guichard:
This is currently up on kdab.searchfox.org using [this Try run](https://treeherder.mozilla.org/jobs?repo=try&revision=36722e667ecbdf459e69861ad735edb9f0677e65). This requires changes to m-c which I have not submitted for review yet, thus marking this as WIP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ that lives in the `shared/` folder in this repository.
      where $REVISION is "mozilla-central.latest".  We extract the specific revision from that.
    - Then https://github.com/mozsearch/mozsearch-mozilla/blob/master/shared/fetch-tc-artifacts.sh
      is invoked and it tries to fetch the result of the taskcluster searchfox jobs
-     for all of our supported platforms (linux64 macosx64 macosx64-aarch64 win64 android-armv7)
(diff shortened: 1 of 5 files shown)


asutherland (changes requested):
Thanks!  A logistical complication is that the change in https://phabricator.services.mozilla.com/D217402 will only update mozilla-central, but this logic is used on beta, release, esr128, and esr115 too.  While to some extent we can perform uplifts of that PR to the other branches, it might be most pragmatic for us to:
- Leave android-armv7 around
- Do add android-aarch64
- Add both android platforms to the logic that prevents us from complaining about missing ios and macosx64-aarch64 jobs.

This will avoid us receiving email warnings for the older branches or ignoring their armv7 builds.

nicolas-guichard:
Ah yes indeed. That should cover both cases then.

arai-a:
can the commit message be reworded to avoid confusion?

nicolas-guichard:
Sure, thanks for reminding me.

asutherland (approved):
Thanks!

PR #261, opened 2024-09-05
Bug 1912078 - Fix wubkat build.

nicolas-guichard:
Searchfox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1912078


wubkat/build
```diff
@@ -53,7 +53,8 @@ cmake $FILES_ROOT \
     -DUSE_LIBBACKTRACE=OFF \
     -DENABLE_BUBBLEWRAP_SANDBOX=OFF \
     -DCMAKE_BUILD_TYPE=Debug \
-    -DPORT=$PORT
+    -DPORT=$PORT \
+    -DUSE_SYSTEM_SYSPROF_CAPTURE=NO
```
emilio, line 57: Hmm, maybe `=OFF` for consistency?

emilio (approved):
If it works this looks good to me, fwiw.

dpino:
The change makes sense, since recently WebKitGTK builds with sysprof by default. The system dependency was added to 'Tools/glib/dependencies/apt'. Probably running 'Tools/gtk/install-dependencies' is another way of fixing this build failure (or manually installing the required dependency).

asutherland:
We do [still run the gtk install-dependencies step in our existing setup script](https://searchfox.org/mozsearch-mozilla/rev/004e82d60ead45e1e722ff503d305469b23c981a/wubkat/setup#79-82):
```
# Support for 22.04 seems to have been officially added in
# https://searchfox.org/wubkat/commit/aa960b1a0c7b980efdbc66718239a449b5966a0c
# so we're now using it directly again.
sudo $FILES_ROOT/Tools/gtk/install-dependencies
```
and I understand that [to pull in the glib deps](https://searchfox.org/mozsearch-mozilla/rev/004e82d60ead45e1e722ff503d305469b23c981a/wubkat/setup#79-82):
```
# Support for 22.04 seems to have been officially added in
# https://searchfox.org/wubkat/commit/aa960b1a0c7b980efdbc66718239a449b5966a0c
# so we're now using it directly again.
sudo $FILES_ROOT/Tools/gtk/install-dependencies
```

Maybe the local testing didn't run these scripts?

wubkat/build
nicolas-guichard: Yes indeed :+1:

asutherland:
Aha, okay, so we can probably re-enable once was move to 24.04.  I have that basically working except the awscli installation situation is weird because on 24.04 the most-supported way to get it is via snap but I believe our base images don't install snap support and I was still looking into that.  (Or more specifically, I thought I had recalled problems from snaps previously, but I think that was just a daemon automatically updating packages in the background and making it effectively impossible to have accurate backtraces if we attached gdb to a hung python instance because the python had been upgraded.)

PR #262, opened 2024-09-30
Fix wubkat build

nicolas-guichard:
I managed a successful indexer run on Friday with this (currently deployed on [kdab.searchfox.org](https://kdab.searchfox.org/wubkat/source/)).

PR #264, opened 2024-10-02
Bug 1912078 - Additional config changes proposed in #262 for config5

asutherland:
There was some additional discussion in
https://github.com/mozsearch/mozsearch-mozilla/pull/262 of other changes we could make.  Since I'm landing a bunch of other updates, I'm making the changes now since I'll potentially need to investigate any fallout.

### DIFF
--- a/wubkat/build
+++ b/wubkat/build
@@ -42,6 +42,19 @@ CXX=${CXX:0:7}
 
 date
 
+# Rationales:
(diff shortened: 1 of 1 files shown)


emilio (approved):

asutherland:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1923545 to track the breakage that may be related to this.  I'm going to hit the revert button now for this in the hopes that developer mode is somehow the problem.  We'll see what happens tomorrow.

PR #267, opened 2024-10-11
Bug 1922408 - update LLVM cmake ignore paths like webkit

asutherland:
Our LLVM semantic build also uses CMake and so the change from CMakeTmp to CMakeScratch is also necessary for LLVM in config4 too.

### DIFF
--- a/config4.json
+++ b/config4.json
@@ -28,7 +28,7 @@
       "github_repo": "https://github.com/llvm/llvm-project",
       "objdir_path": "$WORKING/llvm/objdir",
       "ignore_missing_path_prefixes": [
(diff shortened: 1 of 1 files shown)


asutherland:
It turns out we also need CMakeTmp for some reason; I'll add that:
```
WARN logged_span: tools::file_format::repo_data_ingestion: Problem gathering metadata for /mnt/index-scratch/llvm/objdir/CMakeFiles/CMakeTmp/getErrc.cpp: No such file or directory (os error 2) name="repo_ingestion" uuid=71e60487-b835-4bf7-8ff0-e1166618ef71
```

PR #269, opened 2024-11-10
Bug 1930345 - remove cypress from config4 but retain vcs sync

asutherland:


asutherland:
I'm just going to land this because I did the most minimal approach possible to disabling this tree and it should be obvious tomorrow if I messed things up.

PR #271, opened 2024-11-27
Disable remaining m-c production checks because the HTML check was failing

asutherland:
As noted in the searchfox channel, I think in the face of today's failure on the HTML check and:
- The ergonomics of updating production checks is not great.
- Our test corpus is quite good, including with XPIDL coverage (although this does somewhat depend on data synced from m-c in a manual process), and we haven't really had any significant regressions in quite some time.

It makes sense to just disable the remaining checks here.


asutherland:
gonna retrigger config1 which just failed because of this.

PR #272, opened 2024-11-27
wubkat bustage fix: fix builder name

asutherland:
:padenot noticed the builds are old and it seems it's due to the builder name changing.


asutherland:
The change made the revision happy locally, retriggering config5 which is running right now.

PR #273, opened 2024-12-03
Bug 1900534 - Switch l10n repo to https://github.com/mozilla-l10n/firefox-l10n

asutherland:
I've locally run the steps from
https://github.com/mozsearch/mozsearch/blob/master/docs/newrepo.md for https://github.com/mozilla-l10n/firefox-l10n and created firefox-l10n git and blame lz4 tarballs.  I locally have tested that this works by creating a just-l10n.json config and "build-l10n-repo" target and run with that to verify I didn't mess up anything obvious, but I won't land those changes because this is not a tree that justifies having that; we should just do the dynamic thing so we can specify a config file and a tree in that config file if we want.

### DIFF
--- a/config4.json
+++ b/config4.json
@@ -11,6 +11,8 @@
       "index_path": "$WORKING/l10n",
       "files_path": "$WORKING/l10n/git",
       "git_path": "$WORKING/l10n/git",
+      "git_blame_path": "$WORKING/l10n/blame",
+      "github_repo": "https://github.com/mozilla-l10n/firefox-l10n",
       "objdir_path": "$WORKING/l10n/objdir",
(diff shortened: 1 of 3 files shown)


asutherland:
Note I left the existing tarball around with its existing name, so the existing approach will continue to work until we land this.  While hopefully I didn't mess up the boilerplate for this, for review purposes it's probably worth seeing if people agree with my rationale at https://bugzilla.mozilla.org/show_bug.cgi?id=1900534#c4 as to whether it's fine for existing permalinks to break given the lack of use of the tree (that I could see).

asutherland:
:flod confirmed he's okay with the permalinks breaking, just going to land this.

PR #275, opened 2024-12-10
Bug 1936331 - Remove kaios index

asutherland:


asutherland:
I'm just going to merge this since this is fairly rote/trivial and I'll re-trigger config4 immediately and it should be obvious if I messed up somehow.  Assuming that completes successfully I'll remove the ELB and S3 storages associated.

PR #277, opened 2025-01-25
Bug 1942757 - Stop creating the rust-analyzed file

arai-a:
This is a part of https://bugzilla.mozilla.org/show_bug.cgi?id=1942757

With the https://github.com/mozsearch/mozsearch/pull/860 patch stack, the files in objdir are all reflected to the output.
The rust-analyzed file is created there for historical reason and it's no longer necessary.


asutherland (approved):
Thanks!

PR #278, opened 2025-01-26
Use scip-python from PATH instead of using mise

nicolas-guichard:
This is the counterpart of https://github.com/mozsearch/mozsearch/commit/9ec3c6b0.


asutherland (approved):
Thanks!  It's also fine to merge fixes like these without pre-landing review when it's related to production logistical issues/breakage as long as you're comfortable with the changes and aren't specifically looking for another set of eyes.

PR #279, opened 2025-06-11
Bug 1970807 - Add mozilla-esr140 repo

Eijebong:


asutherland:
Thanks for preparing the patch for this!  Since this is only for mozilla-esr140 and it reuses our shared gecko tarball infrastructure, I don't think there should be any additional steps here beyond me adding the EC2 release-lb ELB ALB rule which I've just done.  (This would not be the case for comm-esr140, which would need manual tarball prep.)

The utc10 config1 run should run soon which should take care of the repo mutations in time for tomorrow and I'll pay attention to any errors.  There is an active release2 indexer currently which will spawn its web-server in a bit, but we shouldn't experience any version skew because we intentionally avoid that.

asutherland:
We should see a fresh config2 run tomorrow... it looks like yesterday's web-server was spawned at 19:16 UTC (after an indexer spawn of 13 UTC)... this is going to make config2 take longer so it could be as late as 21 UTC since I think we're at ~1.75 hours for m-c right now.

PR #311, opened 2026-03-10
Download coverage data for up to 10 mozilla-central pushes

nicolas-guichard:
This removes the logic that “snaps” the indexed revision to the latest with available coverage data and instead tries and downloads coverage data for the last 10 mozilla-central pushes.

The data is retrieved from the Google Storage bucket if available (faster and compressed downloads), or from taskcluster as before.

This is a stop-gap measure while I finish the rebuild-blame equivalent that downloads all coverage data from the bucket to provide a bit of history already and prevent issues such as bug 2021300 by removing the need to parse the task details to map to a mercurial hash.

### DIFF
--- a/firefox-main/setup
+++ b/firefox-main/setup
@@ -4,20 +4,6 @@ set -x # Show commands
 set -eu # Errors/undefined vars are fatal
 set -o pipefail # Check all commands in a pipeline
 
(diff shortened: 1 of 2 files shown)


asutherland (approved):
Fun!  Thank you!

PR #312, opened 2026-03-13
Bug 2023038 - remove cedar while it can't be indexed

asutherland:
The expectation is we will want to index cedar again and once there are usable searchfox artifacts again we can revert this change.


arai-a (approved):
Thank you!

PR #315, opened 2026-04-10
Bug 2013948 - Use codedov2git_gcloud to download historical coverage

nicolas-guichard:
Depends on https://github.com/mozsearch/mozsearch/pull/1068

This downloads `firefox-main-coverage.bundle` from AWS S3 (which currently has 1 year worth of coverage) and adds the last month of coverage if missing before re-uploading.

### DIFF
--- a/firefox-main/setup
+++ b/firefox-main/setup
@@ -21,28 +21,18 @@ $CONFIG_REPO/firefox-shared/checkout-gecko-repos.sh $BUILD_REVISION_TREE "main" 
 $CONFIG_REPO/firefox-shared/fetch-tc-artifacts.sh $BUILD_REVISION_TREE $INDEXED_HG_REV "$PREEXISTING_HG_REV"
 
 echo "Performing setup::codecov2git step for $TREE_NAME : $(date +"%Y-%m-%dT%H:%M:%S%z")"
-mkdir -p "$COVERAGE_ROOT"
+if [ ! -d "$COVERAGE_ROOT" ]; then
+    # Can't directly clone https://searchfox.repositories.s3.amazonaws.com/firefox-main-coverage.bundle because the bucket has got a . in its name.
+    if [[ ! -f "firefox-main-coverage.bundle" ]]; then
+        aws s3 cp s3://searchfox.repositories/firefox-main-coverage.bundle . --no-sign-request
+    fi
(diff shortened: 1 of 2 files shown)


firefox-main/setup
```diff
@@ -21,28 +21,23 @@ $CONFIG_REPO/firefox-shared/checkout-gecko-repos.sh $BUILD_REVISION_TREE "main"
...
-    # First try downloading from the bucket as that's faster, or try the taskcluster API as that's available faster, otherwise bail-out for this push
-    ${CURL} "${COVERAGE_BUCKET_ROOT}/${BUILD_REVISION_TREE}%2F${COVERAGE_HG_REV}%2Fall:all.json.zstd?alt=media" | unzstd > code-coverage-report.json \
-        || ${CURL} ${TC_INDEX_API_URL}/${COVERAGE_INDEX}/artifacts/public/code-coverage-report.json > code-coverage-report.json \
-        || continue
+if [ ! -d "$COVERAGE_ROOT" ]; then
+    # Can't directly clone https://searchfox.repositories.s3.amazonaws.com/firefox-main-coverage.bundle because the bucket has got a . in its name.
+    if [[ ! -f "firefox-main-coverage.bundle" ]]; then
+        aws s3 cp s3://searchfox.repositories/firefox-main-coverage.bundle . --no-sign-request || true
+    fi
+    # Test runs won't have a firefox-main-coverage.bundle
```
asutherland: I don't understand what this is saying.  In general the above aws s3 cp command should always work regardless of whether you have an IAM or not because we grant global read access to the file and we're using `--no-sign-request`.

asutherland (approved):

firefox-main/setup
nicolas-guichard: Hum right this was unclear: my first test indexing runs didn't have a firefox-main-coverage.bundle on the S3 bucket.

My latest test runs do have (1 year of) data on S3 already and all subsequent runs should as well yes.
Would you remove the if (and `|| true` above) and fail if the file is ever removed for some reason?

asutherland: Ah, gotcha, yes it made sense that it wouldn't be there before you uploaded it.

I would lean towards making the file missing being fatal, yeah, since there really isn't any expected situation where we would fail to download it, and it's always frustrating to have silent regressions.

PR #316, opened 2026-04-10
Remove dead coverage.moz.tools links

nicolas-guichard:


asutherland (approved):
Restating:
- mozsearch-mozilla changes corresponding to the removal of the uses in https://github.com/mozsearch/mozsearch/commit/962360f92c4fc559ed1dffc174e317b63097faea from https://github.com/mozsearch/mozsearch/pull/1077

PR #317, opened 2026-05-19
Bug 1923545 - enable VERBOSE=1 for wubkat builds to get more info

asutherland:
Note that when this bug was initially filed I had manually tried to run with VERBOSE=1 and there was no obvious smoking gun, but it's possible the logging situation may have improved in the 2 years since then and/or an LLM might be able to scrape something useful out of the log.


arai-a (approved):
Thank you!

PR #318, opened 2026-05-25
Bug 2042171 - Fix hazard artifact download

padenot:
The browser-haz-debug job now produces .zst artifacts instead of .gz, and allFunctions.txt is no longer uploaded individually (it's only in the hazardIntermediates bundle). Add a download_artifact helper that uses bsdtar so that it works on all common compression formats. Previous file organization is still supported so e.g. Firefox 140 still works.

This requires https://github.com/mozsearch/mozsearch/pull/1089 to be merged to work.

### DIFF
--- a/firefox-shared/fetch-tc-artifacts.sh
+++ b/firefox-shared/fetch-tc-artifacts.sh
@@ -35,6 +35,41 @@ fi
 # ${CURL} url -o outut-location || true
 CURL=${CURL:-"curl -SsfL --compressed"}
 
+warn() { echo "WARNING: $*"; }
+
+# Try suffixes .zst, .gz, and bare; decompress via bsdtar (handles any format).
+# Returns 1 if not found, 2 if decompression failed.
+download_artifact() {
+    local url_base="$1"
+    local outfile="$2"
+    local tmpfile
+    tmpfile=$(mktemp)
+    trap "rm -f '${tmpfile}'" RETURN
+
+    if ${CURL} "${url_base}.zst" -o "${tmpfile}" ||
+       ${CURL} "${url_base}.gz"  -o "${tmpfile}"; then
+        bsdtar -xOf "${tmpfile}" > "${outfile}" || { warn "Failed to decompress ${url_base}"; return 2; }
+    elif ${CURL} "${url_base}" -o "${tmpfile}"; then
+        mv "${tmpfile}" "${outfile}"
+    else
+        return 1
+    fi
+}
+
+# Extract allFunctions.txt from the hazardIntermediates bundle.
+fetch_all_functions_from_bundle() {
+    local haz_url="$1"
+    local tmpfile
+    tmpfile=$(mktemp)
+    trap "rm -f '${tmpfile}'" RETURN
+
+    if ${CURL} "${haz_url}/hazardIntermediates.tar.zst" -o "${tmpfile}" ||
+       ${CURL} "${haz_url}/hazardIntermediates.tar.gz"  -o "${tmpfile}"; then
+        tar xf "${tmpfile}" --no-anchored allFunctions.txt || \
+            warn "Unable to extract allFunctions.txt from hazardIntermediates."
+    fi
+}
+
 # Rewrite REVISION to be a specific revision in case the "latest" pointer changes while
 # we're in the midst of downloading stuff. Using a specific revision id is safer.
 REVISION="${REVISION_TREE}.revision.${INDEXED_HG_REV}"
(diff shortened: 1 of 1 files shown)


asutherland:
It'd be great to file a bugzilla bug for this too and make it depend on bug 2016259.  There are the appropriately hooks in place so if this pull gets the bug number added to it, the PR will automatically be added as an attachment, etc.

padenot:
Ok, this is now in https://bugzilla.mozilla.org/show_bug.cgi?id=2042171 as well.

hotsphink:
Oops, sorry. I forgot about this dependency. Is there a job I can push to try to test whether I'm breaking this? (For future changes, to prevent any further breakage.)

padenot:
I don't think so, but it would be good. I should probably be in searchfox's own tests or health check (if that's a thing on this service).

firefox-shared/fetch-tc-artifacts.sh
```diff
@@ -174,12 +189,19 @@ if [[ -f wpt-manifests.tar.gz ]]; then
...
+# allFunctions.txt may not be uploaded individually; fall back to extracting
+# it from the intermediates bundle in that case.
+if ! download_artifact "${HAZ_URL}/allFunctions.txt" allFunctions.txt; then
+    tmpfile=$(mktemp)
+    if ${CURL} "${HAZ_URL}/hazardIntermediates.tar.zst" -o "${tmpfile}" ||
+       ${CURL} "${HAZ_URL}/hazardIntermediates.tar.gz"  -o "${tmpfile}"; then
+        tar xf "${tmpfile}" --no-anchored allFunctions.txt || true
+    fi
+    rm -f "${tmpfile}"
+    true
```
nicolas-guichard: nit: Does this line do anything?

firefox-shared/fetch-tc-artifacts.sh
```diff
@@ -35,6 +35,23 @@ fi
...
+    local outfile="$2"
+    local tmpfile
+    tmpfile=$(mktemp)
+
+    ${CURL} "${url_base}.zst" -o "${tmpfile}" ||
+    ${CURL} "${url_base}.gz"  -o "${tmpfile}" ||
+    ${CURL} "${url_base}"     -o "${tmpfile}" ||
+    { rm -f "${tmpfile}"; return 1; }
+
+    bsdtar -xOf "${tmpfile}" > "${outfile}" 2>/dev/null || mv "${tmpfile}" "${outfile}"
```
nicolas-guichard: That's a surprising way to “fallback” to `mv` in the non-compressed case. If the extraction fails for some reason the error is ignored and the file will be read as a text file later on.

nicolas-guichard (commented):
The indexer sends an email to the searchfox-aws mailing list when there's a warning in the logfile ([here](https://github.com/mozsearch/mozsearch/blob/master/infrastructure/aws/send-warning-email.py)).
So if you echo "WARNING: …" on download/extract failure we will get notified next time this moves. Alternatively, removing the `|| true` branches would fail the indexer run and keep the previous index online (and also send an email obviously).

asutherland:
The hazard jobs seem to be [tier-1](https://searchfox.org/firefox-main/rev/9d2be0cb8dbf1c3a7ef6ba79be35d637d0327eaa/taskcluster/kinds/hazard/kind.yml#23) so it seems fine for them to cause an indexing failure if we are indexing "firefox-main" (which we can check via `$REVISION_TREE`).

The general problem we've had is that a lot of auxiliary data is very far from tier-1, so having searchfox fail when it's not present wasn't a good heuristic.  (For example [when coverage was broken for 9 months](https://bugzilla.mozilla.org/show_bug.cgi?id=1925873).)

Especially since @arai-a automated the population of the list of trees on the root page and the tree-switcher from the config.json files, it's much more feasible for us to be able to provide better indications of health since we can more easily define what trees are expected to have what functionality/resources available and therefore whether their absence is a problem or not.  And then that can both be surfaced to users and grafana cleanly.

padenot:
In the latest version this warns, as @nicolas-guichard outlined.

asutherland:
Do we have hazard builds on all the firefox-* trees we semantically index?

arai-a:
not all branches have the hazard build.
main, beta, release, esr140, and esr115 has.
So, other branches will hit the warning.

nicolas-guichard (approved):
Thanks!

nicolas-guichard:
The warning was hit:
```
+ curl -SsfL --compressed https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.revision.c9b7663d5238675798b95b93b88e515116db868a.firefox.browser-haz-debug/artifacts/public/build/gcFunctions.txt.zst -o /index/tmp/tmp.O8EFq3vk2H
+ bsdtar -xOf /index/tmp/tmp.O8EFq3vk2H
bsdtar: Error opening archive: Unrecognized archive format
+ warn 'Failed to decompress https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.revision.c9b7663d5238675798b95b93b88e515116db868a.firefox.browser-haz-debug/artifacts/public/build/gcFunctions.txt'
+ echo 'WARNING: Failed to decompress https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.revision.c9b7663d5238675798b95b93b88e515116db868a.firefox.browser-haz-debug/artifacts/public/build/gcFunctions.txt'
WARNING: Failed to decompress https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.revision.c9b7663d5238675798b95b93b88e515116db868a.firefox.browser-haz-debug/artifacts/public/build/gcFunctions.txt
```

As far as I can tell from `man bsdtar`, it requires an (optionally compressed) archive rather than a single compressed file.

PR #319, opened 2026-05-27
Bug 2042171 - Use unzstd or gunzip instead of bsdtar for single files.

nicolas-guichard:
bsdtar only handles (optionally compressed) archives, not single compressed files.

Fixup for https://github.com/mozsearch/mozsearch-mozilla/pull/318


arai-a (approved):
thanks!

PR #320, opened 2026-06-17
Prevent Clang plugin build from trying to overwrite Clang plugin

nicolas-guichard:
Because the Makefile looks at MOZSEARCH_CLANG_PLUGIN_DIR, this would actually overwrite the Clang plugin with the new build. Since the Clang plugin is in the read-only Nix store this now fails.

Don't propagate the env var to the make call.


nicolas-guichard:
I've tested this on kdab.searchfox.org and it fixes the error described by @asutherland in https://github.com/mozsearch/mozsearch/pull/1093#issuecomment-4723986731.

PR #321, opened 2026-06-17
firefox-main: analyze Typescript projects with scip-typescript

nicolas-guichard:
This analyzes all folders which have a tsconfig.json or jsconfig.json with scip-typescript.

### DIFF
--- a/config1.json
+++ b/config1.json
@@ -29,6 +29,10 @@
           "scip_index_path": "$WORKING/firefox-main/python.scip",
           "scip_index_path2": "$WORKING/firefox-main/python2.scip",
           "subtree_root": ""
+        },
+        "typescript": {
+          "scip_index_path": "$WORKING/firefox-main/typescript.scip",
+          "subtree_root": ""
         }
       },
       "group": "Firefox",
--- a/firefox-main/build
+++ b/firefox-main/build
@@ -73,3 +73,14 @@ NODE_OPTIONS="--max-old-space-size=10240" scip-python index --show-progress-rate
 mv python2.scip $INDEX_ROOT
 
 popd
+
+
+echo "Performing build::scip-typescript step for $TREE_NAME : $(date +"%Y-%m-%dT%H:%M:%S%z")"
+pushd "$GIT_ROOT"
+npm install
+# TS_PROJECTS=$(yq -r '.typescript.include | join(" ")' tools/lint/typescript.yml)
+TS_PROJECTS=$(find . -name '[tj]sconfig.json' | grep -v 'node_modules\|third_party')
+scip-typescript index $TS_PROJECTS
+mv index.scip "$INDEX_ROOT/typescript.scip"
+rm -rf node_modules
+popd


nicolas-guichard:
This is up on kdab.searchfox.org. @asutherland I think it would make sense to gather some feedback from impacted users before landing this?

asutherland:
We definitely will want feedback from people, although I think a key ingredient will be to be able to convey what's actually using scip-typescript and what isn't.  A JS subsystem I like to use a lot of the time for looking at JS analysis because I end up using it a lot and it also potentially has a lot of C++ crossings contains [AsyncShutdown.sys.mjs](https://kdab.searchfox.org/firefox-main/rev/bd6e0031ae3274c8f2222d1e7481f0fff43d5eca/toolkit/components/asyncshutdown/AsyncShutdown.sys.mjs) and [nsAsyncShutdown.sys.mjs](https://kdab.searchfox.org/firefox-main/rev/bd6e0031ae3274c8f2222d1e7481f0fff43d5eca/toolkit/components/asyncshutdown/nsAsyncShutdown.sys.mjs) but they were not indexed by scip-typescript (as can be quickly determined by looking at the raw analysis records and the symbol naming scheme.)

To figure out what is scip-typescript, I logged into the server and grepped for "Skipping analysis" in the index-log.  I think there were fewer things there than I expected.  In terms of making the info available to people, maybe we can run a very simple grep-y step on the log that builds a hacky html page that lists all the files and drops it in /pages/ like where we put the settings page.  We could also do something more fancy like put the indexer type in the concise file info and then do a batch file listing page that lists all the files that used that generator.

I found https://kdab.searchfox.org/firefox-main/source/toolkit/components/extensions/ExtensionParent.sys.mjs quite interesting:
- This [Services.catMan.enumerateCategory(CATEGORY_EXTENSION_MODULES)](https://kdab.searchfox.org/firefox-main/rev/bd6e0031ae3274c8f2222d1e7481f0fff43d5eca/toolkit/components/extensions/ExtensionParent.sys.mjs#141) ends up linked to the .d.ts but [on trunk](https://searchfox.org/firefox-main/rev/3b0ec2e6ef5bd362c37420922deaddf9d5d5e175/toolkit/components/extensions/ExtensionParent.sys.mjs#141) we have :arai's possible IDL definitions heuristics which can take us directly to the IDL file which is useful and would be a definite regression to lose.  So I think we'll want to try and add some kind of transformation there.
- Here in [for "load" in lazy.Schemas.load](https://kdab.searchfox.org/firefox-main/rev/bd6e0031ae3274c8f2222d1e7481f0fff43d5eca/toolkit/components/extensions/ExtensionParent.sys.mjs#177) scip-typescript gives us a nice go-to-def whereas [on trunk](https://searchfox.org/firefox-main/rev/3b0ec2e6ef5bd362c37420922deaddf9d5d5e175/toolkit/components/extensions/ExtensionParent.sys.mjs#177) we only have a search.
- But here [for "getStartupChanges" in lazy.AddonManager.getStartupChanges](https://kdab.searchfox.org/firefox-main/rev/bd6e0031ae3274c8f2222d1e7481f0fff43d5eca/toolkit/components/extensions/ExtensionParent.sys.mjs#124) scip-typescript only has a search whereas [on trunk](https://searchfox.org/firefox-main/rev/3b0ec2e6ef5bd362c37420922deaddf9d5d5e175/toolkit/components/extensions/ExtensionParent.sys.mjs#124) js-analyze does have a go-to-def.

nicolas-guichard:
Thanks for taking a look. A first issue was that the whole `toolkit` directory wasn't indexed by scip-typescript. This explains both why `AsyncShutdown.sys.mjs` and `nsAsyncShutdown.sys.mjs` were not indexed, and also why `lazy.AddonManager.getStartupChanges` doesn't link to `toolkit/mozapps/extensions/AddonManager.sys.mjs`. This is solved by https://github.com/mozsearch/scip-typescript/pull/2

I triggered a new indexer run with that fix.

For the IDL issue, IIUC in [idl-analyze.py](https://searchfox.org/mozsearch/source/scripts/idl-analyze.py#351-358) and [webidl-analyze.py](https://searchfox.org/mozsearch/source/scripts/webidl-analyze.py#470), we map to `#enumerateCategory`, and we would need to map to `S_js_typescript_generated/lib.gecko.xpcom.d.ts/global/nsICategoryManager#enumerateCategory().`.
The C++ symbol seems to be retrieved from the C++ analysis (it's not easy to guess either due to mangling), so I think we should do the same for scip-typescript symbols?

asutherland:
Yes, for for the XPIDL idl-analyze.py and WebIDL webidl-analyze.py, it's reasonable and pragmatic to follow the same idiom.

That said, [the proposed semantic linkage enhancement](https://bugzilla.mozilla.org/show_bug.cgi?id=1727789) is also a possibility.  The tl;dr is basically:
- We could make it the crossref stage's problem to do the actual linkage.
- In doing so we can simplify the analyzers, ideally making it feasible for us to have more analyzers in-tree closer to the tooling.  Like the XPIDL and WebIDL code-gens in-tree would be responsible for emitting searchfox analysis data since it's trivial for them to describe the symbol in terms of pretty identifiers.  And ideally that lowers the bar to more Mozilla tooling just generating searchfox data outright.
- There isn't necessarily a major win in these fundamental searchfox use-cases, it's more about the ability to encourage in-tree tool maintainers to be able to help improve searchfox's data without it becoming a searchfox maintainer problem.  It also potentially is a win for being able to define synthetic/higher-level linkages just in [ontology-mapping.toml](https://searchfox.org/mozsearch/source/config_defaults/ontology-mapping.toml), like being able to create synthetic symbols for our nsIObserver(Service) APIs or in your Qt repo for the [signals and slots mechanism](https://doc.qt.io/qt-6/signalsandslots.html).
- This does potentially require crossref to learn to re-write[1] the analysis files for source records for the fancier synthetic linkages.  But in cases where the analyzer introduces completely synthetic symbols where the crossref process can hook those up via binding slots (that is, we don't emit JS symbols directly in the file but instead use binding slot edges), we might not need the rewrite capability.

1: This need not actually be re-writing; we could create auxiliary files that would store a separate record type that decorates the original records.  Or the output-file stage could have to do some additional lookups from the "meta" struct or some other new struct, but that likely would be more wasteful.  Plus there are some other discussed enhancements where I think we definitely do want rewriting like decorating inline SQL strings or cross-referencing spec references in code comments, etc.

asutherland:
Hm, so, I suppose here's an interesting use-case for some semantic linkage support for documentation purposes... I'm thinking out loud here, this is not an explicit request to address this in this PR (but I wouldn't mind if this is something we pursue, as it would be useful).

The sole override mechanism I introduced for C++ use-cases is now firing [here for entries()](https://kdab.searchfox.org/firefox-main/rev/c681e91369f59d0efae43bdc465872b855e8b269/toolkit/components/asyncshutdown/AsyncShutdown.sys.mjs#118)
```js
let entry = this._indirections.entries().next();
```

With a context menu of "Go to definition of Sole Override of [Namespace::entries](https://kdab.searchfox.org/firefox-main/rev/c681e91369f59d0efae43bdc465872b855e8b269/toolkit/components/extensions/Schemas.sys.mjs#3596)" because [we've got this extension of Map in the extensions Schemas module](https://kdab.searchfox.org/firefox-main/rev/c681e91369f59d0efae43bdc465872b855e8b269/toolkit/components/extensions/Schemas.sys.mjs#3252):
```js
class Namespace extends Map {
```

The interesting situation here is because Typescript of course has a typelib for the language builtins, so we have a real symbol for `S_js_typescript_lib/lib.es2015.iterable.d.ts/Map#entries().` but we only have uses for that symbol, nothing that's def/decl/idl/etc for us to go to.  This is a major improvement on [the generic `#entries` symbol search we get from js-analyze](https://searchfox.org/firefox-main/search?q=symbol:%23entries&redirect=false), but the override heuristic is a bit weird there since we don't have the actual def.

But MDN does have a well-defined scheme for built-ins, namely https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries for this case, and so we could define some basic rules for the core typescript libs to map to a new "doclink" edge which could be rendered in the context menu with "def" priority.  We could also hook that up to Java/Android stuff too potentially.

nicolas-guichard:
> it's more about the ability to encourage in-tree tool maintainers to be able to help improve searchfox's data without it becoming a searchfox maintainer problem

The code we're talking about here is generated, right? Could the generators annotate the generated code to link back to the XPIDL/WebIDL source which generated it?

I also have the Java↔C++ use-case in mind. What I did then was to recognize patterns in the generated C++ code to generate annotations to link to the Java symbol, with a caveat for overrides: they need to be in the same order in the source Java file and the generated C++ code. If we're willing to have things in-tree, these annotations could be generated directly by the Java annotation processor, that would be more robust.

asutherland:
> The code we're talking about here is generated, right? Could the generators annotate the generated code to link back to the XPIDL/WebIDL source which generated it?

I don't see why not.  And maybe that could be a net benefit to LLM analysis of the files for local builds?

My perspective has tended to assume we already want to be generating searchfox analysis JSON records directly for the token locations.
 
> I also have the Java↔C++ use-case in mind. What I did then was to recognize patterns in the generated C++ code to generate annotations to link to the Java symbol, with a caveat for overrides: they need to be in the same order in the source Java file and the generated C++ code. If we're willing to have things in-tree, these annotations could be generated directly by the Java annotation processor, that would be more robust.

A disclaimer is that I've discussed the searchfox integration possibilities for XPIDL/WebIDL but I haven't talked to any Android people about doing such a thing, but I don't know of any structural opposition to such a thing.

asutherland:
We seem to be missing https://kdab.searchfox.org/firefox-main/source/mobile/shared/components/geckoview/FilePickerDelegate.sys.mjs and the other files in that directory too.  Those would be quite nice to have.

asutherland (approved):
rs=asuth (rubber-stamp) on any potential follow-up landings or related changes to generally turn these things on and/or mitigate things being on for older branches if they lack info, etc. too.  Thanks!

PR #322, opened 2026-06-26
Bug 2046479 - Add firefox-esr153.

arai-a:
for [bug 2046479](https://bugzilla.mozilla.org/show_bug.cgi?id=2046479)

This does:
  * Add firefox-esr153 tree

### DIFF
--- a/config2.json
+++ b/config2.json
@@ -1,7 +1,7 @@
 {
   "mozsearch_path": "$MOZSEARCH_PATH",
   "config_repo": "$CONFIG_REPO",
-  "instance_type": "t3.2xlarge",
+  "instance_type": "m6id.2xlarge",
 
   "trees": {
     "firefox-beta": {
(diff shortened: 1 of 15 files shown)


asutherland (approved):
Aside: I think we should probably update config2.json to use an `instance_type` of `m6id.2xlarge` just like config1 to make the beta/release/esr trees all have more consistent performance.

Restating:
- This is the first ESR since we added the firefox-main / firefox-* support in https://github.com/mozsearch/mozsearch-mozilla/pull/283
- This patch does more than it strictly needs to because I never cleaned up after the firefox-main migration and it's updating every place that could possibly know about the esr to know about it; we really only need the firefox-* stuff and the config2.json changes.
  - The config1old.json and config2old.json files don't really need to exist anymore, but removing them can be done as a separate bug.  As briefly mentioned in the f-m PR, they were more for rollback-ish purposes.  (Specifically, by having them in the tree, it was possible to trigger an indexer directly against "config1old" while still being able to fix "config1" and from a terminology perspective it would be less ambiguous.)  We can definitely remove these whenever we want.
  - The mozilla-central that exists on config6 does continue to be relevant as long as:
    - The [comm-central tree uses the legacy gecko tarball checkout](https://searchfox.org/mozsearch-mozilla/rev/fd9a0fe72cbcab660e202f2a08b1bf9a497bec87/comm-central/setup#45).  I had some WIP overhauls of the thunderbird branches to be modernized to do semantic stuff that ideally I should pick up again which can moot this need.
    - mozilla-elm still is hosted directly in hg (not git) and exists in [config4.json](https://searchfox.org/mozsearch-mozilla/rev/fd9a0fe72cbcab660e202f2a08b1bf9a497bec87/config4.json).  (We removed cypress and cedar because they were unused/broken.)  mozilla-central is the only job that uploads the updated tarballs so it needs to pull the branches and compute blame to push them so we're not constantly re-fetching stuff.
  - But the mozilla-central scripts strictly speaking don't need to know about esr153, except they do if we ever use the newly introduced mozilla-esr153 for some reason, then we would want those changes.

PR #324, opened 2026-07-07
firefox-main: restore TRYPUSH_REV feature

nicolas-guichard:
Partially reverts 652ce702 which removed the TRYPUSH_REV feature by mistake.


asutherland (approved):
Thanks!

PR #325, opened 2026-07-14
Bug 2054094 - Switch comm-central source to thunderbird-desktop git.

coreycb:
comm-central is now developed in git. Add a fallback path to clone from the new repo and initialize an empty blame repo.

This would require manual removal of S3 tarballs before landing.

### DIFF
--- a/comm-central/setup
+++ b/comm-central/setup
@@ -11,10 +11,15 @@ pushd $INDEX_ROOT
 if [ -d "git" ]
 then
     echo "Found pre-existing comm-central git folder; skipping re-download."
-else
-    wget -nv https://s3-us-west-2.amazonaws.com/searchfox.repositories/comm-central.tar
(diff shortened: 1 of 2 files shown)


asutherland (changes requested):
These blame generation steps generally need to be done manually given the time factors involved (the indexer may time out, etc.).

AFAICT, the repos are not revision compatible so we potentially need an hg double setup like we did for firefox-main to build an additional mapping or all permalinks will break, which we can't do.  Like [this is the last revision in releases-comm-central](https://github.com/mozilla/releases-comm-central/commit/72b8ba0761b3881d926be53f77fbf75d5a9316d5) but if I transplant that commit to the new repo, it's a 404: https://github.com/thunderbird/thunderbird-desktop/commit/72b8ba0761b3881d926be53f77fbf75d5a9316d5

coreycb:
Thanks for taking a look. I'm trying to figure out where the sync for releases-comm-central lives. Hoping we can re-enable that for a bit.

PR #328, opened 2026-08-18
Fix indexing Try runs

nicolas-guichard:
0914e960 broke indexing Try runs because the indexed revision is never an ancestor of $BRANCH.


asutherland (approved):
Ah, right.  And although a check like this could make sense (validate that the indexed revision is on the ancestry line for what we checked out), there's no real need because there's no potential for divergence for try because the indexed revision will also be the revision we are using as the HEAD revision for history.  (We've only decoupled history for non-try branches.)

Thanks!

PR #329, opened 2026-08-21
Build wubkat with Ninja

nicolas-guichard:
Aims to fix this error:
```
CMake Error at Source/cmake/WebKitCommon.cmake:111 (message):
  The GTK port requires the Ninja generator, but this build directory was
  configured with the "Unix Makefiles" generator.

  Re-run CMake with -GNinja or export CMAKE_GENERATOR=Ninja
```

### DIFF
--- a/wubkat/build
+++ b/wubkat/build
@@ -41,6 +41,7 @@ CXX=${CXX:0:7}
 echo "Performing build::cmake step for $TREE_NAME : $(date +"%Y-%m-%dT%H:%M:%S%z")"
 
 cmake $FILES_ROOT \
+    -GNinja \
     -DCMAKE_C_COMPILER="$CC" \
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Thanks!  It sounds like from https://bugs.webkit.org/show_bug.cgi?id=321266 and https://github.com/WebKit/WebKit/pull/71135 that presumably is capturing exactly what we were experiencing, with @arai-a having identified in https://bugzilla.mozilla.org/show_bug.cgi?id=1923545#c13 that there seemed to be dependency problems.

I'll merge this and re-trigger config5/release5.

asutherland:
Ah, I see, probably need https://github.com/mozsearch/mozsearch/pull/1136 too.  I'll merge that and reprovision and then retrigger again.

PR #331, opened 2026-08-28
Re-enable GPU process and match upstream builders more.

asutherland:
Disabling the GPU process is an internal webkit thing and no longer works (there is an unconditional call that depends on it now), plus it may have been just a band-aid for the pre-ninja build dependencies being broken.

While correcting that, I re-evaluated the other things we disabled and it seems like apart from the documentation/introspection generation, we might as well re-enable everything else.  If the build doesn't pass I'll of course re-disable things.

### DIFF
--- a/wubkat/build
+++ b/wubkat/build
@@ -40,21 +40,26 @@ CXX=${CXX:0:7}
 
 echo "Performing build::cmake step for $TREE_NAME : $(date +"%Y-%m-%dT%H:%M:%S%z")"
 
-cmake $FILES_ROOT \
+# The upstream builder we use to determine good revisions has both of the
+# following currently disabled and so we disable them too:
+# -DUSE_LIBBACKTRACE=OFF \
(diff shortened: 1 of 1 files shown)


asutherland:
Experienced a build problem, seems like we can add the deps; I'll double check if there's a way we can be getting those deps out of the tree:
```
-- Found Libseccomp: /usr/lib/x86_64-linux-gnu/libseccomp.so (found version "2.5.5")
CMake Error at Source/cmake/BubblewrapSandboxChecks.cmake:27 (message):
  xdg-dbus-proxy executable not found and is needed for
  ENABLE_BUBBLEWRAP_SANDBOX.  Either install it or use the CMake variable
  DBUS_PROXY_EXECUTABLE to define the runtime path.
Call Stack (most recent call first):
  Source/cmake/OptionsGTK.cmake:499 (include)
  Source/cmake/WebKitCommon.cmake:339 (include)
  CMakeLists.txt:22 (include)
  ```

PR #332, opened 2026-08-28
Update wubkat dependencies and installation method

asutherland:
xdg-dbus-proxy is an explicit configure dependency but not in the upstream package lists.

The install-dependencies script from upstream can take a list of packages and libharfbuzz-dev and gi-docgen are already in its package list so we can remove them.


asutherland:
This seems to have run to completion happily without warnings.  Hopefully we're good for a while now.

PR #333, opened 2026-09-17
Bug 1970809 - Add comm-beta, comm-release, comm-esr140, and comm-esr153

dandarnell:
Now that comm trees have completed their transition to git, I'm opening up this PR to bring over comm-beta, comm-release, comm-esr140, and comm-esr153 in the hopes that they can now be added.

Prior PR: https://github.com/mozsearch/mozsearch-mozilla/pull/282


dandarnell:
@asutherland Would it be possible to get this PR merged to searchfox? Being able to search our ESR trees again, along with monthly and beta, is frequently requested by our development team.

I'm happy to help with any other PRs needed.

asutherland:
This patch as-is won't currently do anything.  Have you tried running searchfox locally on a machine with this patch?

asutherland:
To try and make it easier for you to iterate on this and since ideally you/the Thunderbird team can be an active participant in making sure the Thunderbird indices are happy:
- I've invited you to the [searchfox-aws group which is the source of LDAP ability to do searchfox AWS stuff](https://people.mozilla.org/a/searchfox-aws/).  Once you accept the invite there's info at https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md how to log into the AWS UI for searchfox as well as how to get the CLI AWS stuff going.
- I've pre-created a "thunderbird" channel for thunderbird.searchfox.org to exist for you to do [try-ish runs where you can point an AWS server at your git checkouts/branches and then have it run and stand up a server at thunderbird.searchfox.org that you and others can test again](https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#locally).  Note that these [non-release runs do not run the "upload" scripts](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/infrastructure/aws/index.sh#68-77) in order to avoid experimental changes from potentially contaminating our tarballs, but if you are minting completely new tarball/tree names (more on that below), it could be fine to include a patch in your mozsearch stack that has the "thunderbird" channel do uploads to help with testing, but we would avoid landing that.  In order to get emails about failures when testing you actually want a commit on the "mozsearch" tree anyways.
  - That said, frequently when I've been adding new trees I just run things locally and upload the tarballs from there, but options do include using [trigger_shell.py](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/infrastructure/aws/trigger_shell.py) which sets up an indexer just to be ssh'ed into with it basically stopping after the "setup" phase, but you can do whatever you want on the server once it's set up. This is conceptually similar to [trigger_blame_rebuild.py](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/infrastructure/aws/trigger_blame_rebuild.py) which is supposed to run those build-reblame scripts but I don't know if that mechanism currently works; I know it had broken in the past and I think I tried to fix it when refactoring that python helper script but most of my blame work has required more iteration than that mechanism affords and most developer machines should likely now be much more powerful than the AWS indexers we use.
- Do feel free to upload tarballs to S3 as needed from your local machine for experimentation/etc., and do feel free to create backup copies in S3 is/as-needed.  As long as you're not reusing existing file names in a way that could overwrite other/existing trees and break things it shouldn't be a problem.  One note is that the AWS web UI really hates copying files over 5 GiB and makes you do it locally, but it's possible some of our docs suggests the web UI will work for backing stuff up... it will not in those cases.

What I'm hinting at in the previous message is that this PR currently adds a bunch of trees but with them not being referenced by a config.json file ([config6.json](https://searchfox.org/mozsearch-mozilla/rev/351f1c1704df738522a76cbb728330962e587dc9/config6.json) currently hosts comm-central and comm-esr115 and comm-esr128), nothing would actually index them.  Note that :arai added a very nice import mechanism like you can see in [just-fm.json](https://searchfox.org/mozsearch-mozilla/rev/351f1c1704df738522a76cbb728330962e587dc9/just-fm.json) where you can add an extra config file for test/try runs so if you're iterating you're only waiting on a single tree to run rather than multiple trees.  Especially with mozilla-central currently in config6, that could make things slower than you might like.  Unfortunately mozilla-central does need to stay in config6 for now, but it's fine to remove during testing.  Given the importance of the comm-* trees and especially if you go with adding a bunch of trees rather than creating a shared tree, we should definitely make [config6 be like config1 where the web-server uses an SSD by specifying a fancier instance type](https://searchfox.org/mozsearch-mozilla/rev/351f1c1704df738522a76cbb728330962e587dc9/config1.json#5).  We can add more channels if needed, although for the SSD machines we should be trying to cram as many trees as possible onto them since the SSD eliminates the previous rationale where we were trying to fit the core crossref db and fulltext searches into the 32G of RAM the machine had.

And so one thing I would suggest, if possible, and part of why my own efforts had stalled out because it is potentially more involved if we're dealing with mapping between the old comm-central hg cinnabar revisions and the new trees, is to follow what I did for the [firefox-shared mechanism for the new firefox trees](https://searchfox.org/mozsearch-mozilla/source/firefox-shared) where instead of having a separate tarball for every tree and every tree is its own git repo we instead have a single unified bare repo (which gets checked out as a worktree) plus a single unified bare repo for blame.  We also maintain a bare repo that is based on an old cinnabar checkout of mozilla-central so we have gecko-dev-family old git revisions; this is actually slightly overkill since once we cut over to firefox-main on searchfox we ceased to hand out gecko-dev-family revisions, but it was nice in the sense that it provided more flexibility about testing and the switchover.  But technically we don't really need that old cinnabar tree anymore.  And if you look at [the build-blame CLI docs (which also get automatically integrated into the --help output)](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/tools/src/bin/build-blame.rs#33-107), for helping establish a mapping from old revisions we support both a mapping textfile (which we used for the CVS revisions) plus looking up via cinnabar.  If/when you want to be able to have the comm-central tree cease to exist and instead bounce to the new tree, it then is possible to either just compute such a text-file and avoid the second cinnabar tree or you can use the cinnabar tree.  (The current comm-central tarball is of course the perfect starting point for such a thing.)  Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1890435 and the related commits for more information.

But since my goal here is to unblock you, if you just want to create separate tarballs for each tree as a first step, I'm also fine with that since the main side effect is really just disk usage and also thereby indexing run time.

Please feel free to ask questions in the "searchfox" channel in matrix about things; I am pretty time limited for the time being so I may not be able to respond in depth or anytime soon, and while I cannot speak for the time availability of the other maintainers either, I am sure as a team we can help get you ramped up.  We can also add other TB team members to the LDAP group as needed.

